PR #7, opened 2019-02-04
Update from upstream again

cvium:
Fixes version checks and playback. I'm not sure what webservice.py is used for, but I think the change makes sense for us as we use GUIDs and not int id's (new thing in Emby 4 it seems).

### DIFF
--- a/resources/lib/webservice.py
+++ b/resources/lib/webservice.py
@@ -118,11 +118,11 @@ class requestHandler(BaseHTTPServer.BaseHTTPRequestHandler):
         try:
             params = self.get_params()
 
-            if not params:
+            if not params or params.get('Id') is None:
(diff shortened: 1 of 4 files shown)


resources/lib/webservice.py
```diff
@@ -118,11 +118,11 @@ def do_GET(self):
         try:
             params = self.get_params()
 
-            if not params:
+            if not params or params.get('Id') is None:
                 raise IndexError("Incomplete URL format")
 
-            if not params.get('Id').isdigit():
-                raise IndexError("Incorrect Id format %s" % params.get('Id'))
+            if 'extrafanart' in params['Id']:
```
JustAMan: this change doesn't make sense to me

cvium: It's more a revert than anything.

PR #9, opened 2019-02-04
Re-add dependency add-ons for cleaner paths

cvium:
Don't merge this until the three dependencies are ready.

cc @joshuaboniface

### DIFF
--- a/addon.xml
+++ b/addon.xml
@@ -1,10 +1,13 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <addon  id="plugin.video.jellyfin"
         name="Jellyfin" 
-        version="10.1.0"
-        provider-name="angelblue05">
+        version="10.2.0"
+        provider-name="jellyfin-team,angelblue05">
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
+    <import addon="plugin.video.jellyfin.movies" version="1.0.0"/>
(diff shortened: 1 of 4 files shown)


JustAMan:
@cvium @joshuaboniface please ping this PR when dependencies are up so we can merge it.

anthonylavado:
@cvium @JustAMan What are we waiting on here?

cvium:
We're waiting for @joshuaboniface to fork the `plugin.video.jellyfin.*` deps.

joshuaboniface:
I forgot all about that. I believe we wanted the beta ones correct?

joshuaboniface:
>believe we wanted the beta ones correct? I've got the whole web archive, but where do we need to put them?

@cvium How do we need to proceed here?

oddstr13:
Dependencies are in https://github.com/jellyfin/jellyfin-kodi-plugins, but not yet in the Kodi repo (https://repo.jellyfin.org/releases/client/kodi/).

I am not entirely sure of what exactly this PR or those plugins are fixing, however.

oddstr13:
@cvium What exactly does this PR attempt to solve?
If this separation is beneficial, would it not be better to use "sub-paths" rather than "domains"? So that we chan avoid having those extra dummy dependencies.

oddstr13:
Closing this as there is no progress nor follow-up.
Reopen or just comment if this is still something we would want to do.

PR #11, opened 2019-02-28
Make better install instructions

mcarlton00:
Adding a lot more detail to the install instructions to be more user friendly

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,37 @@ The add-on supports a hybrid approach. You can decide which Jellyfin libraries t
 - and more...
 
 ### Install Jellyfin for Kodi
-1. Install Jellyfin for Kodi from zip.
-2. Within a few seconds you should be prompted for your server-details.
-3. Once you're succesfully authenticated with your Jellyfin server, the initial sync will start. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,14 +39,40 @@ The add-on supports a hybrid approach. You can decide which Jellyfin libraries t
 - and more...
 
 ### Install Jellyfin for Kodi
-1. Install Jellyfin for Kodi from zip.
-2. Within a few seconds you should be prompted for your server-details.
-3. Once you're succesfully authenticated with your Jellyfin server, the initial sync will start. 
-4. The first sync of the Jellyfin server to the local Kodi database may take some time depending on your device and library size.
-5. Once the full sync is done, you can browse your media in Kodi, and syncs will be done automatically in the background.
+
+1. Download the add-on as a [zip file](../../archive/master.zip)
```
cvium: It should be the absolute URL imo. If the readme is moved the link will be incorrect

mcarlton00:
I think we're actually going to move the install guide into readthedocs, and then put a link to them here.  https://github.com/jellyfin/jellyfin-docs/pull/44

I'll come back here with a new PR after we get those docs sorted out.

PR #16, opened 2019-03-20
Update to Kodi MusicDB 72

texke:
Can't find where to put the genres, but people using version 18.1 will be pleased.

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -349,7 +349,7 @@ class Library(threading.Thread):
                 if settings('kodiCompanion.bool'):
 
                     for plugin in self.server['api'].get_plugins():
-                        if plugin['Name'] in ("Jellyfin.Kodi Sync Queue", "Kodi companion"):
+                        if plugin['Name'] in ("Kodi Sync Queue", "Kodi companion"):
                             
                             if not self.fast_sync():
                                 dialog("ok", heading="{jellyfin}", line1=_(33128))
--- a/resources/lib/objects/kodi/music.py
+++ b/resources/lib/objects/kodi/music.py
@@ -189,14 +189,15 @@ class Music(Kodi):
 
         ''' Add genres, but delete current genres first.
         '''
+		#Removed for DB version 72
         if media == 'album':
-            self.cursor.execute(QU.delete_genres_album, (kodi_id,))
-
+            #self.cursor.execute(QU.delete_genres_album, (kodi_id,))
+		
             for genre in genres:
-
+		
                 genre_id = self.get_genre(genre)
-                self.cursor.execute(QU.update_genre_album, (genre_id, kodi_id))
-
+                #self.cursor.execute(QU.update_genre_album, (genre_id, kodi_id))
+		
         elif media == 'song':
             self.cursor.execute(QU.delete_genres_song, (kodi_id,))
 
--- a/resources/lib/objects/kodi/queries_music.py
+++ b/resources/lib/objects/kodi/queries_music.py
@@ -49,11 +49,11 @@ get_album =     		"""	SELECT 	idAlbum
 						"""
 get_album_obj =             [   "{AlbumId}","{Title}","{UniqueId}","{Artists}","album"
                             ]
-get_album_by_name = 	"""	SELECT 	idAlbum, strArtists  
+get_album_by_name = 	"""	SELECT 	idAlbum, strArtistDisp  
 							FROM 	album 
 							WHERE 	strAlbum = ? 
 						"""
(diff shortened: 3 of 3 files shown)


resources/lib/objects/kodi/music.py
```diff
@@ -189,14 +189,15 @@ def add_genres(self, kodi_id, genres, media):
 
         ''' Add genres, but delete current genres first.
         '''
+		#Removed for DB version 72
```
JustAMan: can you use consistent whitespace? here are tabs while the file uses spaces.

resources/lib/objects/kodi/music.py
```diff
@@ -189,14 +189,15 @@ def add_genres(self, kodi_id, genres, media):
...
         if media == 'album':
-            self.cursor.execute(QU.delete_genres_album, (kodi_id,))
-
+            #self.cursor.execute(QU.delete_genres_album, (kodi_id,))
```
JustAMan: can you explain the change?

resources/lib/objects/kodi/queries_music.py
```diff
@@ -82,7 +82,7 @@
     					"""
 add_single_obj =            [   "{AlbumId}","{Genre}","{Year}","single"
                             ]
-add_song =	     		"""	INSERT INTO song(idSong, idAlbum, idPath, strArtists, strGenres, strTitle, iTrack, 
+add_song =	     		"""	INSERT INTO song(idSong, idAlbum, idPath, strArtistDisp, strGenres, strTitle, iTrack,
```
JustAMan: what happens for Kodi < 18.1?

texke: Can your create an objects/Leia branch, like the emby for kodi git?
Seems this is their “solution”.
I’m quite new to kodi addon dev, I presume the next step is having repo’s with different versions.
Did these changes to get jellyfin working on my Kodi, so didn’t think ahead :P

resources/lib/objects/kodi/music.py
texke: Can’t seem to find a substitute for the “genres” table in the new db, seems they store it in the songs table.
Keeping it in the code for reference

resources/lib/objects/kodi/music.py
texke: Will re-submit in the future Leia branch ;)

resources/lib/objects/kodi/queries_music.py
texke: Welp, seems https://github.com/jellyfin/jellyfin-kodi/commit/5fba6555a8ee8e3acbfee7bf79fb571c058bf52b#diff-e986bc05edb125aca769304dee76ab30 had a bunch of code which checked the Kodi version and applied the correct objects patch.
We'll have to bring that back

cvium: Not to dismiss their work, but they are misusing Git branches and we will not be following that model

cvium: This seems much saner: https://forum.kodi.tv/showthread.php?tid=310315

texke: Indeed.
How do you propose to handle multiple versions of the addon itself?

cvium: I have no experience with Kodi plugin development, but I was thinking a branch per Kodi version. Barring any major restructuring in the server API, there shouldn't be many discrepancies between Kodi versions that are non-Kodi version specific (if that makes sense).

JustAMan: Can we just make runtime checks for Kodi versions? Python is very flexible.
This might be easier than maintain two separate branches.

cvium: I'm not a big fan of the Emby model where it automatically updates the 'objects' without consent.

resources/lib/library.py
```diff
@@ -349,7 +349,7 @@ def startup(self):
                 if settings('kodiCompanion.bool'):
 
                     for plugin in self.server['api'].get_plugins():
-                        if plugin['Name'] in ("Jellyfin.Kodi Sync Queue", "Kodi companion"):
+                        if plugin['Name'] in ("Kodi Sync Queue", "Kodi companion"):
```
cvium: This is an unrelated change. It should be made in a separate PR. And I'm also thinking we should remove this check altogether as it's pretty silly. If fast sync is enabled, it should just try to do it and show an appropriate error message if it fails.

texke: Weird, this commit shouldn't be here, I changed this to get rid of the error everytime I try to start Kodi :-)

Issue #15, opened 2019-03-19
Kodi Sync Queue not working currently due to naming mismatch


escabe:
On the Kodi Add-in side, the plugin specifically looks for plugins named "Jellyfin.Kodi Sync Queue" or "Kodi companion" to interact with the Kodi Sync Queue plugin , see:

https://github.com/jellyfin/jellyfin-kodi/blob/d195c4e6d2a2d31242f4a6bd571ac1aab406e1d8/resources/lib/library.py#L352

From the Kodi Sync Queue plugin end, the Add-in currently identifies itself as "Kodi Sync Queue" though (without "Jellyfin." prefix), see:

https://github.com/jellyfin/jellyfin-plugin-kodisyncqueue/blob/8c194349f7ee41c1fb5fd582b8b6392ad66b5fa4/Jellyfin.Plugin.KodiSyncQueue/Plugin.cs#L39

PR #18, opened 2019-04-12
fast sync fix jellyfin/jellyfin-kodi#15

jmeacoe:
…Added "Kodi Sync Queue" to plugins searched when attempting to use fast sync which now matches serverside.

### DIFF
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -349,7 +349,7 @@ class Library(threading.Thread):
                 if settings('kodiCompanion.bool'):
 
                     for plugin in self.server['api'].get_plugins():
-                        if plugin['Name'] in ("Jellyfin.Kodi Sync Queue", "Kodi companion"):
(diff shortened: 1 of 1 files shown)


jmeacoe:
Hey dtorner, I added the space, committed and pushed.  Let me know if there is anything else you need me to do, as this is my first time working with git.

dtorner:
Well, I'm just a user with no admin rights like you, I just jumped in because I had the problem that this pull fixes. I hope an admin will review and approve it in case it is OK to be merged...

cvium:
Honestly, I'd rather the plugin name check wasn't there at all. It should show an error if there's no endpoint listening instead.

jmeacoe:
I agree that there are better potential fixes but as it stands this issue is crippling the usability of the Kodi plugin and negating the jellyfin server plugin completely.
I believe this to be a good fix as a stop gap, and a new issue for improvement can be created, which I should be able to look into it a little later this week, but will take more time as I literally just applied the fix suggested in #15.

cvium:
Approving as a stop-gap

PR #28, opened 2019-05-10
Always use plaintext login

### LINKED ISSUE #22: Password incompatibility with 10.3+
as per: https://www.reddit.com/r/jellyfin/comments/bgmer1/embyforkodi_broken_with_jellyfin_1031031/

Currently, Jellyfin4Kodi sends a prehashed password when you try and authenticate.
The server now requires a salt.

Where the issue is:
https://github.com/jellyfin/jellyfin-kodi/blob/master/resources/lib/jellyfin/core/connection_manager.py#L161

oddstr13:
Fixes #22

This means plaintext passwords now can travel over the internet over http.
You should protect your servers accordingly with https.

### DIFF
--- a/resources/lib/dialogs/loginmanual.py
+++ b/resources/lib/dialogs/loginmanual.py
@@ -120,7 +120,7 @@ class LoginManual(xbmcgui.WindowXMLDialog):
 
         mode = self.connect_manager['server-mode']
         server = self.connect_manager['server-address']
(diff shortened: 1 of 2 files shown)


mcarlton00:
Working here.  Just purged one of my kodi boxes and re-installed with the branch.  Everything re-synced and plays fine.

PR #29, opened 2019-05-14
Link to readthedocs for install instructions

mcarlton00:
Just changing the install instructions to point to the readthedocs client page instead of maintaining 2 copies of the same instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,8 @@ The add-on supports a hybrid approach. You can decide which Jellyfin libraries t
 - and more...
 
 ### Install Jellyfin for Kodi
-1. Install Jellyfin for Kodi from zip.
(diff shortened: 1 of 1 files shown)


EraYaN:
Wouldn't the git specific stuff be out of scope since it's the same for every git repo? We should probably link to some git doc page for that.

mcarlton00:
I'm inclined to think that anybody who's going far enough to pull a specific branch or PR is going to have a good idea of how to do that already, or at the very least be able figure it out easily.  I'm not strongly opposed to the idea, but it does seem a bit unnecessary.

As for the initial question of installing from git, we could leave a simplified set of instructions in place and point out that the "recommended" way of installing is via the repo with a link to readthedocs.

PR #38, opened 2019-07-08
Greatly increase the amount of logging information.

oddstr13:
The main change so far is to include file paths and line numbers on log output.

Log all calls to `__getitem__` and `__shortcuts__` (`DEBUG`, noisy)

Some flake8 errors corrected.

Currently prints some stacktraces that probably shouldn't be printed, but my opinion on that is to keep it like that until exception handling can be improved (and made much more specific) in those areas.
```log
00:16:39.614 T:140707650131712  NOTICE: JELLYFIN.database.jellyfin_db -> ERROR::resources/lib/database/jellyfin_db.py:161 'NoneType' object has no attribute '__getitem__'
                                              File "resources/lib/database/jellyfin_db.py", line 159, in get_version
                                                version = self.cursor.fetchone()[0]
                                            TypeError: 'NoneType' object has no attribute '__getitem__'
```

Also added editorconfig.

### DIFF
--- a/resources/lib/jellyfin/core/api.py
+++ b/resources/lib/jellyfin/core/api.py
@@ -334,6 +334,3 @@ class API(object):
         return self._delete("Videos/ActiveEncodings", params={
             'DeviceId': device_id
         })
-
-    def delete_item(self, item_id):
-        return self.items("/%s" % item_id, "DELETE")
--- a/resources/lib/jellyfin/core/configuration.py
+++ b/resources/lib/jellyfin/core/configuration.py
@@ -16,6 +16,7 @@ LOG = logging.getLogger('JELLYFIN.'+__name__)
 
 #################################################################################################
 
+
 class Config(object):
 
     def __init__(self):
@@ -25,6 +26,7 @@ class Config(object):
         self.http()
 
     def __shortcuts__(self, key):
+        LOG.debug("__shortcuts__(%r)", key)
 
         if key == "auth":
             return self.auth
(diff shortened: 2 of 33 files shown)


resources/lib/player.py
```diff
@@ -1,447 +1,447 @@
-# -*- coding: utf-8 -*-
```
oddstr13: It seems this file had Windows file endings. Long live editorconfig?

resources/lib/jellyfin/core/api.py
```diff
@@ -334,6 +334,3 @@ def close_transcode(self, device_id):
         return self._delete("Videos/ActiveEncodings", params={
             'DeviceId': device_id
         })
-
-    def delete_item(self, item_id):
-        return self.items("/%s" % item_id, "DELETE")
```
oddstr13: flake8 was kind enough to informing me that this was a duplicate definition.
You should be able to find the exact same code further up.

resources/lib/jellyfin/core/configuration.py
```diff
@@ -38,14 +40,16 @@ def __shortcuts__(self, key):
...
         return self.data.get(key, self.__shortcuts__(key))
 
     def __setitem__(self, key, value):
         self.data[key] = value
 
     def app(self, name, version, device_name, device_id, capabilities=None, device_pixel_ratio=None):
-        
         LOG.info("Begin app constructor.")
+        # import traceback
+        # LOG.debug(''.join(['\n'] + traceback.format_stack()))
```
cvium: Remove or uncomment

oddstr13: Right, that's from attempting to trace down what's happening with the `config['app.name']` being `None` exception.
I didn't want to lose it, so I commented it out in stead of removing. Fixed now.

PR #40, opened 2019-07-11
Bump version for 0.2.0

oddstr13:


addon.xml
```diff
@@ -27,14 +27,16 @@
   <extension point="xbmc.addon.metadata">
     <platform>all</platform>
     <language>en</language>
-    <license>GNU GENERAL PUBLIC LICENSE. Version 2, June 1991</license>
+    <license>GNU GENERAL PUBLIC LICENSE. Version 3, 29 June 2007</license>
```
joshuaboniface: This can't be done unless its explicitly "or later".

oddstr13: Just for keeping the information in one place; I simply updated this value to the current license from the repo. The license file change was done upstream back in january.

PR #48, opened 2019-07-22
Add Simplified Chinese translation

biaji:


biaji:
> I don't read chinese, so it is a little hard for me to review the translations here.
> 
> I do however see that some strings are missing. Does anyone know how fallback is handled, if at all?
> 
> I also see that the word "artwork" is copied literally.
> Artwork means a work of art, in this context DVD box art, movie poster or similar - a image / picture / photo.
> Surely there is a better translation for that than using the english word?

Yes, there do have some fuzzy strings I don't know how to deal with. 
The word 'artwork'. We don't have a single word to cover the meaning of it. I can translate it as "艺术品", but it's wired for chinese to express it in this way.   So I think keep it as english maybe the best way.

joshuaboniface:
Weblate for this repo is up now.

PR #54, opened 2019-08-18
Make sure __base__ is the first place we look for packages.

oddstr13:
`resources/lib/entrypoint/service.py` is importing `setup`, which just so happens to be a system package on this computer, which exits on import.

This was leading to a hard to track down (debug prints all the way from entry point) bug with no error message - it just didn't work.

These changes make sure the local addon files are searched first for modules to import, instead of last.


oddstr13:
Ugh, my local master is tracking origin not upstream, I'll try to get the rebase done tomorrow

PR #55, opened 2019-08-24
Updates logging to be more consistent

graybush:
The way the logging was initialized varied from file to file. This just
updates the init to be the same everywhere. Also updates some LOG.warn
to be LOG.info.


oddstr13:
Diff is a bit large to review on mobile, but it looks like there are more changes here than described.

resources/lib/client.py
```diff
@@ -82,12 +83,7 @@ def get_device_id(reset=False):
     if client_id:
         return client_id
 
-    directory = xbmc.translatePath('special://profile/addon_data/plugin.video.jellyfin/').decode('utf-8')
-
-    if not xbmcvfs.exists(directory):
```
cvium: Why was this removed?

resources/lib/database/__init__.py
```diff
@@ -211,50 +212,37 @@ def reset():
...
-
-    if dialog("yesno", heading="{jellyfin}", line1=_(33087)):
-
-        xbmcvfs.delete(os.path.join(addon_data, "settings.xml"))
-        xbmcvfs.delete(os.path.join(addon_data, "data.json"))
-        LOG.info("[ reset settings ]")
-
-    if xbmcvfs.exists(os.path.join(addon_data, "sync.json")):
-        xbmcvfs.delete(os.path.join(addon_data, "sync.json"))
+    if window('jellyfin_sync.bool'):
```
cvium: What is this change for?

resources/lib/database/__init__.py
```diff
@@ -346,49 +329,34 @@ def get_sync():
...
-        xbmcvfs.mkdirs(path)
-
     try:
-        with open(os.path.join(path, 'data.json')) as infile:
+        if os.path.exists(os.path.join(ADDON_DATA, 'data.json')):
+            pass
+        elif os.path.exists(os.path.join(ADDON_DATA, 'data.txt')):
+            os.rename(os.path.join(ADDON_DATA, 'data.txt'), os.path.join(ADDON_DATA, 'data.json'))
+        else:
+            open(os.path.join(ADDON_DATA, 'data.json'), 'a')
```
cvium: This doesn't seem right. What is it supposed to do?

service.py
```diff
@@ -68,29 +70,32 @@ def run(self):
...
+    LOG.info("Delay startup by %s seconds.", DELAY)
+
+    if not xbmcvfs.exists(ADDON_DATA):
+        xbmcvfs.mkdirs(ADDON_DATA)
 
     while True:
 
-        if not settings('enableAddon.bool'):
-            LOG.warn("Jellyfin for Kodi is not enabled.")
+        if settings('enableAddon.bool'):
```
cvium: Why invert the if? It seems only the indentation level has been increased, so I don't think it was necessary

graybush:
So this was just supposed to be just the logging stuff. I didn't realize
that since I had not created a branch additional changesets added to master
would automatically get added to this pull request. I will create a branch
for future changesets.

oddstr13:
I've added a bunch of logging recently. Most of the exceptions aren't actually crashes, just over-eager logging of except blocks. Some have been nececary to track down underlying issues (catching Exception everywhere hides many bugs).

We're now printing a bunch of stack traces on caught exceptions, to make debuging things easier. As such, a stack trace may not equal a bug.

Issue #42, opened 2019-07-14
Remove requests and dateutil from the repo.


oddstr13:
Remove requests and dateutil from the repo and pull them in as dependencies from the Kodi repo in stead.

PR #57, opened 2019-08-30
Removed bundled libraries

TrueTechy:
Removed the bundled libraries from the addon and using the ones imported from the kodi repo instead as per issue #42


addon.xml
```diff
@@ -5,6 +5,8 @@
         provider-name="Jellyfin Contributors, angelblue05">
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
+    <import addon="script.module.requests" version="2.9.1"/>
```
mcarlton00: Can we update requests to a more recent version?  Looks like latest is 2.22.0

addon.xml
```diff
@@ -5,6 +5,9 @@
         provider-name="Jellyfin Contributors, angelblue05">
   <requires>
     <import addon="xbmc.python" version="2.25.0"/>
+    <import addon="script.module.requests" version="2.22.1"/>
```
mcarlton00:
```suggestion
    <import addon="script.module.requests" version="2.22.0"/>
```
Typo here.  2.22.1 doesn't exist yet, time traveller

addon.xml
```diff
@@ -5,6 +5,9 @@
...
+    <import addon="script.module.dateutil" version="2.8.0"/>
```
mcarlton00:
```suggestion
    <import addon="script.module.dateutil" version="2.7.3"/>
```
Maybe this is why the full source code was in the repo instead of importing it, but the latest version of the library available for kodi is 2.7.3.  https://github.com/Razzeee/script.module.dateutil
As far as I can tell, going down to 2.7.3 doesn't hurt anything.  I was able to sync my library and playback files with the older version just fine.  We could also request upstream update to a newer version

addon.xml
```diff
@@ -5,6 +5,9 @@
...
+     <import addon="script.module.web-pdb" />
```
mcarlton00: Left your debugging line here.  Not sure if intentional

PR #62, opened 2019-09-01
Updates project structure to Kodi v17

graybush:
Moves pngs from the root to the resources directory and updates references to said pngs in the source code.

I was unsure about `kodi_icon.png`. The only reference I could find is possibly in line 316:
https://github.com/jellyfin/jellyfin-kodi/blob/31aa2073ad84cd0bc0a61963b9517bbb9c0747ab/resources/lib/monitor.py#L299-L317


dkanada:
Actually, I recently deprecated the icon URL as part of my effort to remove any reliance on remote content. The icons were only used for the web client so I added new icons to the web source, which will pull them straight off the web server like any other file, so that specific line can be removed.

graybush:
> Actually, I recently deprecated the icon URL as part of my effort to remove any reliance on remote content. The icons were only used for the web client so I added new icons to the web source, which will pull them straight off the web server like any other file, so that specific line can be removed.

Remove the line from the file and the file itself from the repo or just the line?

mcarlton00:
Remove the `IconUrl` and the image file.  It might get reimplemented [here](https://github.com/jellyfin/jellyfin-web/blob/1131327edee4c97b260b1483bef1fe26332b2ab0/src/scripts/imagehelper.js#L4) in the jellyfin-web repo later.

Issue #27, opened 2019-05-10
Remove `__shortcuts__` and clean up related code.


oddstr13:
Going via  `__shortcuts__` and `__getitem__` makes it harder to figure out where functions are used.

https://github.com/jellyfin/jellyfin-kodi/blob/c12918f32112b09593cbf5dd415b6e5c9b62895b/resources/lib/jellyfin/core/connection_manager.py#L62-L97

TrueTechy:
Adding this list here to keep track of work still needing done 

### **Shortcuts**
resources/lib/jellyfin/core/configuration.py

- [ ] auth 
- [ ] app 
- [ ] http 
- [ ] data

resources/lib/jellyfin/core/connection_manager.py

- [ ] clear 
- [ ] servers 
- [ ] reconnect/refresh 
- [ ] login
- [ ] server 
- [ ] server-id 
- [ ] server-version 
- [ ] user-id
- [ ] public-users 
- [ ] token
- [ ] manual-server
- [ ] connect-to-server 
- [ ] server-address 
- [ ] revoke-token
- [ ] server-mode

resources/lib/jellyfin/core/ws_client.py

- [ ] config 
- [ ] websocket 
- [ ] callback 
- [ ] auth
- [ ] api
- [ ] connected   

### Getitem
resources/lib/jellyfin/__init__.py:
- Has a decorator function on set/get functions, might need to reimplement these elsewhere

resources/lib/jellyfin/client.py:
- [ ]  config
- [x]  http
- [ ]  websocket
- [ ] callback
- [ ] auth
- [x] api
- [ ] connected

resources/lib/jellyfin/core/configuration.py: 
- [ ] \_\_shortcut\_\_

resources/lib/jellyfin/core/connection_manager.py: 
- [ ] \_\_shortcut\_\_

resources/lib/objects/movies.py:
- [ ] Movie
- [ ] BoxSet
- [ ] UserData
- [ ] Removed

resources/lib/objects/music.py:
- [ ] MusicArtist/AlbumArtist
- [ ] MusicAlbum
- [ ] Audio
- [ ] UserData
- [ ] Removed

resources/lib/objects/musicvideos.py:
- [ ] MusicVideo
- [ ] UserData
- [ ] Removed

resources/lib/objects/tvshows.py:
- [ ] Series
- [ ] Season
- [ ] Episode
- [ ] UserData
- [ ] Removed

TrueTechy:
[...]

PR #64, opened 2019-09-03
Remove __shortcuts__ in http.py and some in client.py  #27

TrueTechy:
Removed http and api from getitem in resources/lib/jellyfin/client.py as well as all shortcuts inresources/lib/jellyfin/client.py as part of the cleanup effort for #27


cvium:
Quite an undertaking. Looks interesting. Good luck!

PR #66, opened 2019-09-07
Cleanup __shortcuts__ in client.py and associated code

TrueTechy:
Related to the cleanup as per #27


resources/lib/jellyfin/core/http.py
```diff
@@ -100,7 +100,7 @@ def request(self, data, session=None):
                     continue
 
                 LOG.error(error)
-                self.client['callback']("ServerUnreachable", {'ServerId': self.config['auth.server-id']})
+                self.client.callback("ServerUnreachable", {'ServerId': self.config.data['auth.server-id']})
```
mcarlton00: `server-id` was replaced with `server_id` in multiple places in other files above.  Most of the project uses the underscore version, so I think we should finish these re-namings so everything is consistent.

resources/lib/jellyfin/core/http.py
```diff
@@ -122,12 +122,12 @@ def request(self, data, session=None):
                 if r.status_code == 401:
 
                     if 'X-Application-Error-Code' in r.headers:
-                        self.client['callback']("AccessRestricted", {'ServerId': self.config['auth.server-id']})
+                        self.client.callback("AccessRestricted", {'ServerId': self.config.data['auth.server-id']})
```
mcarlton00: rename `server-id` with `server_id` here

resources/lib/jellyfin/core/http.py
```diff
@@ -122,12 +122,12 @@ def request(self, data, session=None):
...
 
                         raise HTTPException("AccessRestricted", error)
                     else:
-                        self.client['callback']("Unauthorized", {'ServerId': self.config['auth.server-id']})
-                        self.client['auth/revoke-token']
+                        self.client.callback("Unauthorized", {'ServerId': self.config.data['auth.server-id']})
```
mcarlton00: rename `server-id` with `server_id` here

mcarlton00:
Requested changes were a misunderstanding, there's 2 different things in play with a similar name (to be changed at a later date).

PR #67, opened 2019-09-08
Cleanup getitems

TrueTechy:
remove the __getitem__ functions from remaining files as per #27 

__init__.py was bypassed by directly referencing functions which was done in previous commits

for the media files, none of the shortcuts seem to be triggered and no other references to them could be found in the code so they were removed, testing shows no adverse effects


TrueTechy:
Closed due to complete messup with the rebase. Created new PR

Issue #23, opened 2019-05-01
Improvement: remove for loop in plugin checking process


jmeacoe:
Currently the application checks to see if the server has the "kodi sync queue" plugin installed by polling through all the plugins on the server and matching by name.

https://github.com/jellyfin/jellyfin-kodi/blob/c12918f32112b09593cbf5dd415b6e5c9b62895b/resources/lib/library.py#L349-L363

Some potential solutions:
1. Remove the checking for the plugin, the application already has catches in place for failure, but would be much messier and wouldn't have the friendly pop up message for users informing they need the plugin installed.
2. Create an api call in the plugin, that can verify installation.
3. Change check for plugin at setting initialization, so it is only done when settings are changed, and not at startup.

oddstr13:
As far as I can tell, this is only ran once at start — resulting in a single API request — and only when the "Kodi companion" is enabled in the options.

jmeacoe:
This is in relation to PR: #18 
The issue is more dependent on the name checking of the plugin.  I've meant to look into it myself, but haven't had time this week.

PR #71, opened 2019-09-12
Change companion detection to endpoint check #23

TrueTechy:


resources/lib/library.py
```diff
@@ -348,8 +348,7 @@ def startup(self):
             if settings('SyncInstallRunDone.bool'):
                 if settings('kodiCompanion.bool'):
 
-                    for plugin in self.server.jellyfin.get_plugins():
-                        if plugin['Name'] in ("Jellyfin.Kodi Sync Queue", "Kodi companion", "Kodi Sync Queue"):
+                    if self.server.jellyfin.check_companion_installed():
 
                             if not self.fast_sync():
```
cvium: Fix indentation

PR #89, opened 2019-09-27
Remove unnecessary code from server connection

TrueTechy:
The function didn't do much that was useful and was at worst breaking basic auth if the username or password had a capital letter since it was making them all lowercase


TrueTechy:
Given the get item fiasco I wanted to do more thorough testing before removing it completely, at least if this goes through it'll fix the auth issue

Issue #97, opened 2019-09-29
Unable to fetch the libraries


drissslim:
**Describe the bug**
Using the 0.3.0 version of the plugin with a fresh installation. I'm not able to fetch the libraries. This worked fine before with the 0.2.1

**To Reproduce**
You need to start using a fresh install and not updated version of the jellyfin kodi plugin.

**Expected behavior**
Fetch all libraries when selecting all. But no library is displayed.

**Logs**

**Screenshots**

**System (please complete the following information):**
 - OS: Android TV
 - Jellyfin Version: 10.3.7
 - Kodi Version: 18.3
 - Addon Version: 0.3.0

**Additional context**

mcarlton00:
0.3.2 was pushed out yesterday, can you try updating and see if the problem still exists?

drissslim:
I'm still getting the same issue. Even after reinstalling the plugin in version 0.3.2

mcarlton00:
I've received a similar report in Matrix for an Nvidia Shield and was just able to replicate it on my HTC (Running Android 7).  Authenticated to the server, but unable to browse/sync libraries to the local device.

Logs from montery on Matrix:
https://drive.google.com/file/d/1kIhSvhA5UJHSH6ixBA3Peg9iZnqpWgVw/view?usp=sharing

bwaddington:
Describe the bug
Using the 0.3.2 version of the plugin. I'm not able to update the libraries. 

Expected behavior
upate all libraries

System (please complete the following information):

    OS: windows
    Jellyfin Version: 10.3.7
    Kodi Version: 18.4
    Addon Version: 0.3.2

    OS: dietpi
    Jellyfin Version: 10.3.7
    Kodi Version: 18.2
    Addon Version: 0.3.2

TrueTechy:
@bwaddington can you post your kodi.log file after you've tried to update the libraries?

bwaddington:
[kodi.log](https://github.com/jellyfin/jellyfin-kodi/files/3668226/kodi.log)

aoc:
I'm experiencing a similar problem. I updated to 0.3.2, performed a local database reset to start from scratch.

Eventually I can see and play the media if I go to `Add-ons -> Video Addons -> Movies (dynamic)`. But Kodi's `Movies` and `TV Shows` sections remain empty.

I see some errors in the log but I'm not sure if they're related:
http://paste.kodi.tv/ivazanokiz

aoc:
[...]

PR #98, opened 2019-09-29
Add explicit file encoding to file handlers

TrueTechy:
As part of one of the log file in #97 there was exceptions noting the system couldn't determine the default encoding thus they're now coded explicitly

>  ERROR: Exception in thread Thread-18:
>                                             Traceback (most recent call last):
>                                               File "/home/jenkins/workspace/Android-ARM64-Leia/tools/depends/xbmc-depends/aarch64-linux-android-21-release/lib/python2.7/threading.py", line 801, in __bootstrap_inner
>                                               File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/plugin.video.jellyfin/resources/lib/library.py", line 101, in run
>                                                 if not self.startup():
>                                               File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/plugin.video.jellyfin/resources/lib/library.py", line 325, in startup
>                                                 Views().get_views()
>                                               File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/plugin.video.jellyfin/resources/lib/views.py", line 221, in get_views
>                                                 save_sync(self.sync)
>                                               File "/storage/emulated/0/Android/data/org.xbmc.kodi/files/.kodi/addons/plugin.video.jellyfin/resources/lib/database/__init__.py", line 357, in save_sync
>                                                 with open(os.path.join(path, 'sync.json'), 'w') as outfile:
>                                             IOError: could not determine default encoding

### DIFF
--- a/resources/lib/database/__init__.py
+++ b/resources/lib/database/__init__.py
@@ -333,7 +333,7 @@ def get_sync():
         xbmcvfs.mkdirs(path)
 
     try:
-        with open(os.path.join(path, 'sync.json')) as infile:
+        with open(os.path.join(path, 'sync.json'), encoding='utf-8') as infile:
             sync = json.load(infile)
     except Exception:
         sync = {}
@@ -354,7 +354,7 @@ def save_sync(sync):
 
     sync['Date'] = datetime.datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ')
 
-    with open(os.path.join(path, 'sync.json'), 'w') as outfile:
+    with open(os.path.join(path, 'sync.json'), 'w', encoding='utf-8') as outfile:
         data = json.dumps(sync, sort_keys=True, indent=4, ensure_ascii=False)
         outfile.write(unicode(data))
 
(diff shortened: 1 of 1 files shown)


oddstr13:
`encoding` is a Python 3 specific keyword argument to `open()`.
This will break for Kodi 17 and Kodi 18.

TrueTechy:
> `encoding` is a Python 3 specific keyword argument to `open()`.
> This will break for Kodi 17 and Kodi 18.

We're using the python library IO which works in both 2 and 3. The docs say python 2 has the encoding options in the [docs]( https://docs.python.org/2/library/io.html)

oddstr13:
Ah, that would be why then. I searched the codebase for more instances of open, but those aren't using the io library.

Should those other `open()` calls also use the IO library then, with the `encoding` argument?

TrueTechy:
We only changed the DB file to IO as a solution to support unicode characters in the server name as that was breaking the addon using non-ascii characters in their server name. We could change them over to IO and specify the encoding we just haven't gotten to reviewing those parts of the code yet

PR #101, opened 2019-10-02
Remove version check code, queries and language data

TrueTechy:
Reviewing the emby for kodi repo and forum shows that this code was added once back before jellyfin was forked and hasn't been touched since. It looks like it was used to correct a problem where users would need to reset their databases when upgrading their server. However since the version data is never set in the database this query always returns None and never actually executes the rest of the code just throwing an exception every time

In the kodi music object file there are references to check version but this is implemented in that file and will have to be removed before the table can also be removed from the database


mcarlton00:
I'm not 100% sure we want to take this out.  We've seen a bit of weirdness going from server 10.3.7 to 10.4, and there's a good chance it'll happen again with 10.5 in the future.  It may be better to leave it in and actually finish implementing it.

Thoughts?

TrueTechy:
I feel like as we work out the kinks and weirdness in the addon and understand it better to improve it we should hopefully be able to stop this break in compatability between server versions. I just felt forcing the reset (which can rejected by the user mind you thus not really 100% fixing the problem) isn't the best way to handle this

resources/lib/database/queries.py
```diff
@@ -1,4 +1,5 @@
 
+<<<<<<< HEAD
```
oddstr13: This is an artefact from the merge conflict.
Personally I would've used the file from master, and just repeated the changes - This file only had a few variables removed IIRC, which is rather quick to repeat

TrueTechy: Damn, thought I got them all, shame it can't just do my changes over the top of the current master

oddstr13: Oh, you did a merge - have a look into `git rebase` - it essentially takes your changes, and re-applies them on top of the changes from another branch - wouldn't have changed the fact that me touching all the files caused a merge conflict tho.

TrueTechy: I tried that first, all it said was "noop"

PR #103, opened 2019-10-02
Enclose address input in square brackets if it is an IPv6 address

### LINKED ISSUE #79: Fails to connect to IPv6 addresses directly
**Describe the bug**
Addon cannot connect to server if given address is an IPv6 address

**To Reproduce**

1. Enter ::1 in host section
2. Click connect

**Expected behavior**
Addon connects to server

**Logs**
`InvalidURL: Failed to parse: http://::1:9898/emby/system/info/public`

 - OS: Ubuntu
 - Jellyfin Version: 10.4
 - Kodi Version: 18.3
 - Addon Version: 0.2.1

**Additional context**
Seem to be due to the request not properly encapsulating the IPv6 address since it has a port on it as well. Should be enclosed by square brackets [https://softwareengineering.stackexchange.com/questions/311413/how-should-we-represent-an-ipv6-address-with-port-number-in-text](https://softwareengineering.stackexchange.com/questions/311413/how-should-we-represent-an-ipv6-address-with-port-number-in-text)

oddstr13:
Fixes #79 

Input `field.getText()` returns a bytestring rather than a unicode string for some reason.

### DIFF
--- a/resources/lib/dialogs/servermanual.py
+++ b/resources/lib/dialogs/servermanual.py
@@ -4,6 +4,7 @@
 
 import logging
 import os
+import ipaddress
 
 import xbmcgui
 import xbmcaddon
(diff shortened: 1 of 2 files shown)


TrueTechy:
Preliminary testing shows this fixes #79

PR #109, opened 2019-10-05
Remove dependency on `ipaddress` (not stdlib in py2)

### LINKED ISSUE #108: Can't run on LibreElec
**Describe the bug**
The addon will install, but fail to run when installing the latest version on LibreElec

**To Reproduce**
1. Install addon
2. After installing, it errors

**Logs**
[kodi.log](https://github.com/jellyfin/jellyfin-kodi/files/3692737/kodi.log)

 - OS: LibreElec 9.0.2
 - Jellyfin Version: Any
 - Kodi Version: 18.2
 - Addon Version: 0.3.3

**Additional context**
Seems to be a regression from #103.  The `ipaddress` module doesn't appear to be available on LibreElec, and because of the design of the system I haven't found a way to install it.

```
LivingRoom:~ # python
Python 2.7.15 (default, May  4 2019, 18:38:15)
[GCC 8.2.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ipaddress
  File "<stdin>", line 1, in <module>
ImportError: No module named ipaddress
```

oddstr13:
Let's copypaste from StackOverflow instead!
[![Wait, forgot to escape a space.  Wheeeeee[taptaptap]eeeeee.](https://imgs.xkcd.com/comics/regular_expressions.png)](https://www.xkcd.com/208/)

Also, `localhost:///8096/emby/...` is not a valid HTTP(S) URL.

Fixes #108

### DIFF
--- a/resources/lib/dialogs/servermanual.py
+++ b/resources/lib/dialogs/servermanual.py
@@ -4,7 +4,7 @@
 
 import logging
 import os
-import ipaddress
+import re
 
 import xbmcgui
 import xbmcaddon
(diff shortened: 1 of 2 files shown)


weaselshit:
Why not use [socket.inet_pton()](https://docs.python.org/2/library/socket.html#socket.inet_pton)?

Here's a small demo:
```
>>> socket.inet_pton(socket.AF_INET6,'fe80::3:2:1')
b'\xfe\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x03\x00\x02\x00\x01'
>>> socket.inet_pton(socket.AF_INET6,'fe80::3:2::1')
  File "<stdin>", line 1, in <module>
OSError: illegal IP address string passed to inet_pton
```

oddstr13:
Mainly because I wasn't aware of it.

But looking at the docs linked, it says availability: Unix, so it is platform dependent wether this function is available or not.

PR #117, opened 2019-10-06
Add support for upnext addon

### LINKED ISSUE #83: Support for NextUp Service
Please add support for NextUp service, in short; it shows a popup at the end of the episode with info about the next one and allows you to play it instantly
As described:
https://github.com/im85288/service.upnext/wiki/Addon-Integration

Currently clicking it does nothing.

mcarlton00:
Finishes integration with the upnext addon.  Most of the groundwork was already there, but this check was failing and I added extra log messages.  Fixes #83

### DIFF
--- a/resources/lib/monitor.py
+++ b/resources/lib/monitor.py
@@ -66,7 +66,7 @@ class Monitor(xbmc.Monitor):
 
         elif sender.startswith('upnextprovider'):
             LOG.info('Attempting to play the next episode via upnext')
-            method = '.'.join(method.split('.')[1:])
+            method = method.split('.', 1)[1]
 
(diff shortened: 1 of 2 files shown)


resources/lib/monitor.py
```diff
@@ -65,9 +65,11 @@ def onNotification(self, sender, method, data):
             data = json.loads(data)[0]
 
         elif sender.startswith('upnextprovider'):
-            method = method.split('.')[1]
+            LOG.info('Attempting to play the next episode via upnext')
+            method = '.'.join(method.split('.')[1:])
```
JustAMan:
```suggestion
            method = method.split('.', 1)[1]
```
this is slightly faster :)

PR #124, opened 2019-10-11
Update library management dialog titles

### LINKED ISSUE #121: Popups for managing libraries all have the same title
**Describe the bug**
All the popups which are used to manage the libraries after setup all have the same title

**To Reproduce**
1. Within the addon, Go to manage libraries
2. Click on any of the library options that isn't add libraries
3. Note that all titles say "Select the libraries to add"

**Expected behavior**
Titles reflect the action being taken

 - OS: Any
 - Jellyfin Version: Any
 - Kodi Version: 18.3 and 18.4 tested
 - Addon Version: 0.3.4

TrueTechy:
Fixes #121

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -933,3 +933,15 @@ msgstr ""
 msgctxt "#33197"
 msgid "A sync is already running, please wait until it completes and try again."
 msgstr ""
+
+msgctxt "#33198"
+msgid "Select the libraries to update"
+msgstr ""
+
+msgctxt "#33199"
+msgid "Select the libraries to repair"
+msgstr ""
+
+msgctxt "#33200"
+msgid "Select the libraries to remove"
+msgstr ""
--- a/resources/lib/library.py
+++ b/resources/lib/library.py
@@ -493,7 +493,16 @@ class Library(threading.Thread):
 
         choices = [x['Name'] for x in libraries]
         choices.insert(0, _(33121))
-        selection = dialog("multi", _(33120), choices)
+
+        titles = {
+            "RepairLibrarySelection": 33199,
+            "SyncLibrarySelection": 33198,
+            "RemoveLibrarySelection": 33200,
+            "AddLibrarySelection": 33120
+        }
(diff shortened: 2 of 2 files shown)


resources/language/resource.language.en_gb/strings.po
```diff
@@ -933,3 +933,15 @@ msgstr ""
 msgctxt "#33197"
 msgid "A sync is already running, please wait until it completes and try again."
 msgstr ""
+
+msgctxt "#33198"
+msgid "Select the libraries to update"
+msgstr ""
+
+msgctxt "#33199"
+msgid "Select the libraries to repair"
```
dkanada: Is it required to add translations in the wrong language to all files, or will the client fallback to English when a translation doesn't exist?

TrueTechy: I'm sure it would use english if the translation doesn't exists. I haven't used it before but I believe weblate will see them in the other language files and makes them available for translation in the interface

dkanada: If it will fallback to English I would say to just not include the additions to the other language files and let Weblate handle that automatically, but either way is fine.

TrueTechy:
I wasn't sure if we late would be able to determine from the English string file that the new ones should be in the others. Also what python would do if a user is using a language that hasn't been translated thus no listing for it would be found in the string file, I think it only defaults to English because we're using the Msgid wrong

resources/language/resource.language.en_gb/strings.po
TrueTechy: I wasn't sure if we late would be able to determine from the English string file that the new ones should be in the others. Also what python would do if a user is using a language that hasn't been translated thus no listing for it would be found in the string file, I think it only defaults to English because we're using the Msgid wrong

dkanada: On Android it's common practice to leave untranslated strings out of the file, but I will defer to you since I know nothing about how Kodi handles translations.

TrueTechy: After doing some testing, it seems Kodi just displays the English text if no translation is found in the other files. I have thus removed them

PR #128, opened 2019-10-29
Remove default connection port

mcarlton00:
Removes default port mapping of 8096 (discussed in matrix)

### DIFF
--- a/resources/language/resource.language.de_de/strings.po
+++ b/resources/language/resource.language.de_de/strings.po
@@ -56,10 +56,6 @@ msgctxt "#30024"
 msgid "Username"
 msgstr "Benutzername"
 
-msgctxt "#30030"
-msgid "Port number"
-msgstr "Portnummer"
-
 msgctxt "#30091"
 msgid "Confirm file deletion"
 msgstr "Löschen von Dateien bestätigen"
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -50,10 +50,6 @@ msgctxt "#30024"
 msgid "Username"
 msgstr ""
 
-msgctxt "#30030"
-msgid "Port number"
-msgstr ""
-
 msgctxt "#30091"
 msgid "Confirm file deletion"
 msgstr ""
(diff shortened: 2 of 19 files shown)


dkanada:
Technically the goal is to remove the port field entirey and just use a hostname field for the entire URL instead.

mcarlton00:
so the idea is to just have something like `192.168.1.10:8096` or `jellyfin.example.com:8096`?  That's going to add a bit of confusion once the baseurl stuff lands.  Some users will inevitably be trying `example.com/jellyfin:8096` and wondering why things aren't working properly.  What's the end goal of not having a port field at all?

oddstr13:
Currently I have to explicitly clear the port field in order to add my server which is under a sub-path behind reverse proxy. I'd prefer a URL field pointing to the Jellyfin base path

mcarlton00:
Personally, I'd rather the ability to leave the port field blank and default to 80/443 internally depending on if https is available.  That way there's sane defaults for the majority of users, but if there's a good reason to override it's easy to do so.  I just really don't like mixing URLs/ports in the same field.  It gets messy and especially once we have working subpaths will likely lead to a lot of user confusion

oddstr13:
Well, I think a "copy paste" from the URL you'd use for the web interface would be easiest for the end user.

Non-default port means that the port is part of the URL.

On October 31, 2019 10:05:53 PM UTC, mcarlton00 <notifications@github.com> wrote:
>Personally, I'd rather the ability to leave the port field blank and
>default to 80/443 internally depending on if https is available.  That
>way there's sane defaults for the majority of users, but if there's a
>good reason to override it's easy to do so.  I just really don't like
>mixing URLs/ports in the same field.  It gets messy and especially once
>we have working subpaths will likely lead to a lot of user confusion
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/jellyfin/jellyfin-kodi/pull/128#issuecomment-548588002

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mcarlton00:
@oddstr13 @dkanada that should take care of it.  Working fine against my servers, but could use testing from somebody else before we merge to verify

TrueTechy:
Tested on desktop. Syncs and plays fine

PR #131, opened 2019-11-04
Bump version for 0.3.5

mcarlton00:


oddstr13:
I'm having trouble doing a db reset and full sync against 10.4.1 - may or may not be related to the changes since last time I checked out master - more testing needed, as multiple things changed.

mcarlton00:
Is it related to the baseurl change in the server?  I found that I had to do a full reset and redo the server info/user login before data would start flowing again.

mcarlton00:
We'll merge this after #132 lands.  Might as well get that fix in this release as well since it's right on the edge of being ready

mcarlton00:
ew, that got ugly.  closing and opening a new clean one

PR #132, opened 2019-11-04
Fix libraries sync when jellyfin library has apostrophes in name

### LINKED ISSUE #87: Verify addon works with apostrophes in library names
**Describe the bug**
Mentioned by /u/Quixventure on reddit.

"Also, for library issues, I found that if I had an apostrophe (like "Kid's Movies") in the name of my share in Jellyfin, the Kodi client would sync all the data and it would be accessible, but the virtual source would appear empty. You could see stuff in "movies" but not in the Jellyfin add-on>Kid's Movies"

"It was a while ago, and I just renamed my folders, but it could be an issue causing others to see nothing depending on how they are configured."

 - Jellyfin Version: any?
 - Kodi Version: any?
 - Addon Version: any?

loidy:
Fixes #87

### DIFF
--- a/resources/lib/helper/utils.py
+++ b/resources/lib/helper/utils.py
@@ -293,9 +293,9 @@ def indent(elem, level=0):
 def write_xml(content, file):
     with open(file, 'w') as infile:
 
+        # replace apostrophes with double quotes only in xml keys, not texts
         def replace_apostrophes(match):
             return match.group(0).replace("'", '"')
-
         content = re.sub("<(.*?)>", replace_apostrophes, content)
 
         content = content.replace('?>', ' standalone="yes" ?>', 1)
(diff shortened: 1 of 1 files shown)


dkanada:
@mcarlton00 maybe merge this before #131

loidy:
> Not sure this is the best solution since it changes the names on disk, but as long as it works this seems like a good bug fix. I'd add a comment above the method name to explain why you added this method.

I wanted to retain previous functionality (replacing apostrophes with double quotes), I don't know why it is there, so I just added regex substitution to replace apostrophes only in element keys and not texts.

mcarlton00:
Confirmed that it fixes the library browsing through the addon menu (seems to need to restart kodi to take effect though).  I'll rebase my release PR after this is merged.  

Changing the files on disk seems like the most sane approach at this time, since otherwise you'd be changing values dynamically when you read from the file every startup.  

I do agree that a comment would be nice.  Just something like "This replaces ' with " when .....".  It'll make the whole codebase easier to maintain in the future if some of the oddities are documented in the code.

PR #140, opened 2019-11-05
Remove WOL Code

TrueTechy:
Removes code relating to Wake On Lan functionality that was never fully implemented and thus done nothing in it's current state

### DIFF
--- a/resources/lib/jellyfin/core/connection_manager.py
+++ b/resources/lib/jellyfin/core/connection_manager.py
@@ -174,8 +174,6 @@ class ConnectionManager(object):
         if CONNECTION_MODE['Local'] not in tests:
             tests.append(CONNECTION_MODE['Local'])
 
-        # TODO: begin to wake server
-
         LOG.info("beginning connection tests")
         return self._test_next_connection_mode(tests, 0, server, options)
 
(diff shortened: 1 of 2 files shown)


resources/lib/jellyfin/core/api.py
```diff
@@ -64,12 +61,7 @@ def sessions(self, handler="", action="GET", params=None, json=None):
         else:
             return self._get("Sessions%s" % handler, params)
 
-    def users(self, handler="", action="GET", params=None, json=None):
-        if action == "POST":
-            return self._post("Users/{UserId}%s" % handler, json, params)
-        elif action == "DELETE":
-            return self._delete("Users/{UserId}%s" % handler, params)
-        else:
+    def users(self, handler="", action="GET", params=None, json=None): ## Remove action & JSON, maybe others
```
oddstr13: These changes are potentially breaking (`POST` and `DELETE` are in use), and unrelated to wake on lan.

resources/lib/jellyfin/core/api.py
```diff
@@ -53,9 +53,6 @@ def _delete(self, handler, params=None):
 
     #################################################################################################
 
-    def try_server(self):
-        return self._get("System/Info/Public")
```
oddstr13: Isn't used anywhere as far as I can tell, but it's still a change that is unrelated to WoL

PR #142, opened 2019-11-06
Handle unicode characters without using the io library

### LINKED ISSUE #115: Cannot sync videos on Xbox one, prompted to manually add server on restart
**Describe the bug**
As per #97 I'm experiencing similar behavior on Xbox One. 
Every time I restart jellyfin plugin or the Kodi app I am prompted to manually add server, I login and select server, 
I am asked to sync empty shows etc but do not see a prompt to sync all shows as I do on my other devices. 
I have gone to manage libraries and update libraries. 
I only have an All option, no TV shows or Movies option. I press okay and have no further progress from there.
I see no synced videos and am prompted to repeat the above every Kodi restart.

 - OS: Xbox One
 - Jellyfin Version: 10.3.7
 - Kodi Version: 18.4
 - Addon Version: 0.3.4

**Additional context**
I tested jellyfin server running off my MacBook last weekend on macos, installed Kodi on my Xbox one at that point with this plugin and had everything working at that point. 

Have since moved to running the jellyfin server on my Linux server on fedora 30. 
I have tested Kodi on multiple other machines with this plugin and cannot reproduce this behavior.
I'll try to get some logs off the Xbox and upload them later.

oddstr13:
Fixes #115

The reason for `UNICODE = type(u"")` in place of just `unicode` is forward compat with Py3 - if there's better ways to do this, please do enlighten me :)

### DIFF
--- a/resources/lib/database/__init__.py
+++ b/resources/lib/database/__init__.py
@@ -7,7 +7,6 @@ import logging
 import json
 import os
 import sqlite3
-from io import open
 
 import xbmc
 import xbmcvfs
@@ -21,6 +20,9 @@ from objects import obj
 LOG = logging.getLogger("JELLYFIN." + __name__)
 
 #################################################################################################
+
+
+UNICODE = type(u"")
 
 
 class Database(object):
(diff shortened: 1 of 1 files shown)


oddstr13:
`UNICODE` should be swapped with `six.text_type` once we start the Py3 porting proper
https://six.readthedocs.io/#six.text_type

PR #145, opened 2019-11-11
Removed connection tests and minor cleanup

### LINKED ISSUE #77: Remove connection tests
Currently the addon tests to determine which kind of connection mode is needed for the server. Given that it should work for either kind using the same method this code should be removed to simplify the process and ease maintainability

TrueTechy:
Removed the connection tests, and thus removed the functions which returned the appropriate address after changing the two places the addresses are set from their specific names to the generic 'address'

resolves #77

### DIFF
--- a/jellyfin_kodi/monitor.py
+++ b/jellyfin_kodi/monitor.py
@@ -144,9 +144,8 @@ class Monitor(xbmc.Monitor):
             self.void_responder(data, item)
 
         elif method == 'GetServerAddress':
-
-            server_data = server.auth.get_server_info(server.auth.server_id)
-            server_address = server.auth.get_server_address(server_data, server_data['LastConnectionMode'])
+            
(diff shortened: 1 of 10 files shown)


jellyfin_kodi/monitor.py
```diff
@@ -144,9 +144,8 @@ def onNotification(self, sender, method, data):
             self.void_responder(data, item)
 
         elif method == 'GetServerAddress':
-
-            server_data = server.auth.get_server_info(server.auth.server_id)
-            server_address = server.auth.get_server_address(server_data, server_data['LastConnectionMode'])
+
```
mcarlton00: Nit-picky extra new line here.  No idea why it was there in the old version

TrueTechy: They are all like that so I just kept it the same as the rest

PR #154, opened 2019-12-08
Change string Kodi Companion to Kodi Sync Queue

bjoernfan:
Case insensitive.


bjoernfan:
I think this works for most languages, except the non-latin ones. Not sure if the exceptions should be changed to empty strings as the translation is invalid -- otoh I guess the string lookup won't work anyway.

Found this as I was looking for Kodi Companion for Jellyfin, which... yeah, didn't find much under that name. :)

PR #160, opened 2019-12-13
Correct filepath when entering data to database

### LINKED ISSUE #53: Incorrect TV Show Metadata/File Issue
I have added my TV and Movie libraries to Kodi using this (pretty awesome) addon, but I am having a problem with a few TV shows, not all of them.

Example of the problem:
I navigate in Kodi to a TV show and go to information for an episode. It shows the correct thumbnail and title for the episode, but shows the plot and actors for a movie in my library. When I play the episode, it actually plays the movie (that the metadata was from). As far as I can tell this issue only effects some of my TV shows (Planet Earth, South Park) and does not appear to affect movies. Some shows will also not show any information at all (which is my "default select" action in Kodi making it very frustrating to play these videos). The metadata in Jellyfin web interface is correct. 

My environment:
Kodi - Windows 10 PC, Latest version of Kodi
Jellyfin - Docker container, Latest version
Addon - Latest version installed from Repo
Kodi Sync Queue plugin is installed

I will attempt to get some screenshots demonstrating this strange occurrance. Any help would be appreciated. It does look to be posibly related to this previously unresolved issue: https://github.com/jellyfin/jellyfin-kodi/issues/24#issue-440057406

mcarlton00:
Fixes #53 

Without including the libraryId in the path, it just kept overwriting the top level with whatever library had most recently synced

### DIFF
--- a/jellyfin_kodi/objects/movies.py
+++ b/jellyfin_kodi/objects/movies.py
@@ -175,7 +175,7 @@ class Movies(KodiDb):
             obj['Path'] = obj['Path'].replace(obj['Filename'], "")
 
         else:
-            obj['Path'] = "plugin://plugin.video.jellyfin/"
+            obj['Path'] = "plugin://plugin.video.jellyfin/%s/" % obj['LibraryId']
             params = {
                 'filename': obj['Filename'].encode('utf-8'),
                 'id': obj['Id'],
--- a/jellyfin_kodi/objects/musicvideos.py
+++ b/jellyfin_kodi/objects/musicvideos.py
@@ -163,7 +163,7 @@ class MusicVideos(KodiDb):
             obj['Path'] = obj['Path'].replace(obj['Filename'], "")
 
         else:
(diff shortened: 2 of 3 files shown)


cvium:
What about the other library types?

mcarlton00:
As far as I can tell, it's only an issue with the movies and tv shows libraries.  When I was syncing music it had no effect, but I've never tried music videos or mixed content

cvium:
Isn't it better to keep it consistent though?

mcarlton00:
Generally yes.  I was only focusing on movies/shows to address the known issue.  I'm not sure if the other libraries do the same thing or not, I'll need to check on them later after work today/this weekend

mcarlton00:
Music is in it's own database that's formatted different and is fine, and I've updated Music Videos as well just for consistency's sake

LittleCannon:
I've just applied changes to 2 existing installations and everything is OK now.
No more overwritten metadata.
Well done!

PR #167, opened 2019-12-19
Removed unnecessary log messages and changed some to debug

TrueTechy:


jellyfin_kodi/jellyfin/http.py
```diff
@@ -198,7 +199,7 @@ def _get_header(self, data):
                 'Content-type': "application/json",
                 'Accept-Charset': "UTF-8,*",
                 'Accept-encoding': "gzip",
-                'User-Agent': self.config.data['http.user_agent'] or "%s/%s" % (self.config.data['app.name'], self.config.data['app.version'])
+                'User-Agent': self.config.data['http.user_agent'] or "%s/%s" % (self.config.data.get('app.name', 'Jellyfin for Kodi'), self.config.data.get('app.version', "0.0.0.0"))
```
mcarlton00: why '0.0.0.0'?  We only do 3 levels deep in our versioning.

TrueTechy: Too used to writing default route IP addresses?

TrueTechy:
I'm kinda hoping to do a complete redo of the configuration management in the addon since as it stands we're getting a ton of issues where things are set in some config objects but not others and thus you get all the errors when trying to do things like the http requests which require data often not set in these config objects. From what I've seen all these functions that are doing anything useful will throw errors when something goes wrong so it'd be safe to assume all was fine in them if nothing appears?

At that point shouldn't we fix why the UI is being weird. Not convinced we want the average user to be looking in the log files to know what's going on with the addon

mcarlton00:
That's fine, if the config management is being redone we can revisit the logs there.  With these changes we're basically adding 2 more files to the "not logged at all" pile, and I'm not comfortable reducing our visibility of program flow anymore than the codebase itself does.

For the record, I haven't seen the UI being weird during sync, but I like having the messages there in the logs to review regardless so I can easily see where the library is at.  And I don't want to be forced to run in debug mode and sort through irrelevant messages just for a status check.

TrueTechy:
Client.py now has no logging but I'd say those two log lines aren't very useful for an external view as they relate only to how the server identifies Kodi. 

I'd of said the UI does a reasonable Job at displaying sync progress and depending on the library having it throwing hundreds if not thousands of lines into it when doing a sync isn't necessary in regular use, hence it only being required when debugging. If the UI isn't providing an adequate status check then maybe that's something we need to look at separately?

jellyfin_kodi/client.py
```diff
@@ -96,15 +96,15 @@ def get_device_id(reset=False):
...
+        LOG.debug("Generating a new GUID.")
 
         client_id = str("%012X" % create_id())
         file_guid = xbmcvfs.File(jellyfin_guid, 'w')
         file_guid.write(client_id)
 
     file_guid.close()
-
-    LOG.info("DeviceId loaded: %s", client_id)
+
```
dkanada:
```suggestion

```

jellyfin_kodi/jellyfin/http.py
```diff
@@ -147,7 +147,8 @@ def request(self, data, session=None):
                 raise HTTPException(r.status_code, error)
 
             except requests.exceptions.MissingSchema as error:
-                raise HTTPException("MissingSchema", {'Id': self.config.data['auth.server']})
+                LOG.error("Request missing Schema. "+str(error))
```
dkanada:
```suggestion
                LOG.error("Request missing Schema. " + str(error))
```

jellyfin_kodi/monitor.py
```diff
@@ -104,21 +104,21 @@ def onNotification(self, sender, method, data):
 
         try:
             if not data.get('ServerId'):
-                raise Exception("ServerId undefined.")
-
-            if method != 'LoadServer' and data['ServerId'] not in self.servers:
+                server = Jellyfin()
```
dkanada: Why remove the exception here?

TrueTechy: Because that exception throws stuff into the logs which users seem to think is the cause of every issue they have with Kodi but really is just a bad way to handle parameters being empty 

example: https://github.com/jellyfin/jellyfin-kodi/issues/83#issuecomment-534892009

jellyfin_kodi/client.py
TrueTechy: I'm not entirely sure what this suggestion means

mcarlton00: there's an extra newline that doesn't need to be there

TrueTechy: Ah it was there before so I just left it in

PR #188, opened 2020-02-12
Fix kodi 19 websocket control

### LINKED ISSUE #183: controlling session from browser not working
**Describe the bug**
I have a jellyfin 10.4.3, i've a kodi 19 connected with the the git version of the addons, if i go to the webpage of jellyfin i can see the session in kodi, i can see what it's playing but i can not interact with it, no pause, skip or even adding element to the queue, there is any configuration that i'm missing (on server or kodi) ? 

i've also the kodi queue sync plugin enabled in jellyfin

FYI
to install the git addons in kodi 19 i had some issue with module kodi-six, i think i have an older version and it was not able to find it, after some try on removing the module and adding it again, and try to update it i was able to install it

**Logs**
i'm not sure is a kodi addons issue because i can not see any error in the kodi.log file

 - OS: Ubuntu
 - Jellyfin Version: 10.4.3
 - Kodi Version: 19
 - Addon Version: 0.4.0 git

mcarlton00:
Add encode/decode shenanigans to make sure string/byte types match across python 2 and python 3.  Fixes remote control from the Web UI in Kodi 19 installs.

Closes #183

### DIFF
--- a/jellyfin_kodi/jellyfin/websocket.py
+++ b/jellyfin_kodi/jellyfin/websocket.py
@@ -545,8 +545,8 @@ class WebSocket(object):
             return False
         result = result.lower()
 
-        value = key + "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
-        hashed = base64.encodestring(hashlib.sha1(value).digest()).strip().lower()
+        value = key + "258EAFA5-E914-47DA-95CA-C5AB0DC85B11".encode()
+        hashed = base64.encodestring(hashlib.sha1(value).digest()).strip().lower().decode()
         return hashed == result
 
     def _read_headers(self):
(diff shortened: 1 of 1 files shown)


jellyfin_kodi/jellyfin/websocket.py
```diff
@@ -545,8 +545,8 @@ def _validate_header(self, headers, key):
             return False
         result = result.lower()
 
-        value = key + "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
-        hashed = base64.encodestring(hashlib.sha1(value).digest()).strip().lower()
+        value = key + "258EAFA5-E914-47DA-95CA-C5AB0DC85B11".encode()
```
dkanada: What exactly is this string for?

mcarlton00: Hippy space magic?  At this point i genuinely don't know.

It's generating a uuid to use as a key (line 236), then combining that key with this string and hashing it, and somehow that's equivalent to what the websocket returns when the authentication/handshake process is successful (lines 516-527).  I'm not going to pretend to know how that string was found at this point.

PR #191, opened 2020-02-14
Translates German to be less formal

NeroBurner:
Most of the German translations use the informal "du" instead of the formal "Sie".
This patch unifies the usage by removing the few formal occurences


resources/language/resource.language.de_de/strings.po
```diff
@@ -1026,21 +1026,21 @@ msgstr "Aktiviere Jellyfin für Kodi"
...
 msgid "Select the libraries to update"
-msgstr "Markieren Sie die zu aktualisierenden Bibliotheken"
+msgstr "Markiere die zu aktualisierenden Bibliotheken"
 
 msgctxt "#33197"
 msgid "A sync is already running, please wait until it completes and try again."
 msgstr ""
-"Die Synchronisation läuft bereits, warten Sie bitte bis es fertig ist und "
-"versuchen sie es später erneut."
+"Die Synchronisation läuft bereits, warte bitte bis es fertig ist und "
```
AndreasGB: bis sie

NeroBurner: Maybe also "Eine Synchronisation" (--> "A sync") instead of "Die Synchronisation" ("The sync")?

I'd replace "es" with "diese"

"Eine Synchronisation läuft bereits, warte bitte bis diese fertig ist und "

AndreasGB: Even better :)

mcarlton00:
Thanks for the help.  I'll trust Andreas about phrasing.  My only real request is to post your translations through Weblate [here](https://translate.jellyfin.org/projects/jellyfin/jellyfin-kodi/de/) instead.  We haven't 100% confirmed this, but it seems like when we change the translations directly in git it will sometimes break the link to Weblate and throw errors later.

NeroBurner:
@mcarlton00 I tried but the search doesn't work for me and I don't want to click forward 222 times to get the translations in. Any tips on how to work with the translation system?

NeroBurner:
merged with weblate https://github.com/jellyfin/jellyfin-kodi/commit/406f9bbab21b0215758a630d5262ce9f306b0618

PR #193, opened 2020-02-16
Replaced debug output of json.dumps with an indirection to lazy __str__.

druscoe:
json.dumps is a processing intensive operation. This is being called
every time data is received from the server (most noticeably during
library updates) for debug logging. If the user has debug logging
disabled (the default option) then the user is still paying for
processing which is discarded.

The fix is to add a level of indirection where the dumps function is
only called if a string representation of the json is requested; ie.
when the debug string is evaluated.


mcarlton00:
I'm not seeing any noticeable speed difference during syncs.  Can you explain how these changes are supposed to work?

druscoe:
> I'm not seeing any noticeable speed difference during syncs.

I use kodi on lower powered ARM boxes. Every cycle wasted there is very noticeable.
To emulate on my dev machine I hacked the server fetch limit up to a few thousand.
On my ARM box I am witnessing ~3x increase in performance when importing a library.

> Can you explain how these changes are supposed to work?

It simply avoids calling json.dumps(..., indent=4) unless debug output is enabled. It does this by deferring that call until the string representation of the JsonDebugOutput is requested. The string representation is not requested if logging.level >= INFO and so the call is avoided.

https://pythonprogramming.net/__str__-__repr__-intermediate-python-tutorial/
__str__ is called whenever the interpreter gets:
str(obj)
x = "my string: %s" % obj
print(obj)

The original method:
log.debug(json.dumps(data, indent=4))
will always convert the python data structure into a string. Why waste this processing?

druscoe:
As an example of the ineffenciency of json pretty printing:

```
>>> import timeit
>>> timeit.timeit("import json; d=dict((a,a) for a in xrange(9)); json.dumps(d)")
10.93553900718689
>>> timeit.timeit("import json; d=dict((a,a) for a in xrange(9)); json.dumps(d, indent=4)")
41.399051904678345
```

We should only take this performance hit if we want debug output. The cost of pretty printing json can be reduced to zero for any regular user.

mcarlton00:
Just for future reference, that's the kind of information that's super helpful in the comment when opening the PR.  Or opening an issue beforehand and referencing that in the PR.  Just for future maintainability, it makes it way easier to understand why things were done in a certain way.

On my desktop, I'm seeing a difference of ~2 seconds on a library of ~1000 movies and ~100 collections.  On a Raspberry Pi 3 against the same libraries, I'm seeing a difference of 11 seconds.   Which I would say is less than significant, but the initial syncing process is fairly awful on those arm devices, so every bit helps.  And obviously I've only tested against movies, while shows are the much bigger slog to get through.

My main concern is increasing code complexity without noticeable benefit.  Does this need to be it's own class/file combination?  It seems like it could be moved to a function (or stay as a class if needed) under `jellyfin_kodi/helper/utils.py` instead of adding another file.  Possibly move to `jellyfin_kodi/helper/loghandler.py` as well, but if I'm understanding correctly then I'm fairly confident it won't help solve the problem here.

Along the same maintainability line, whether it stays in it's own file or moves somewhere else, please put a comment or two in the code explaining why it exists.  Just as simple as "Prevents unneeded calls to `json.dumps` when logging" or something along those lines is plenty for me.  Not on every log line, but certainly on the function/class definition.  I realize a lot of things in the codebase aren't commented well right now, but the goal is to improve things as we go.

druscoe:
I don't want to appear condescending by explaining the __obvious__ when submitting patches. If my reasoning is not explained too well, or needs further explanation, then I am always happy to expand. If you want a documentation trail for this project then I will be happy to give fuller explanations.

The reason I started digging into this code was purely performance; I could not feasibly use this plugin as-is so I'll improve for myself... and others :) 
I am seeing a much larger increase than yourself but every little helps as you say. I want to keep digging in the performance area as I ideally want much more fluid kodi experience. I moved to jellyfin as I was under the impression that the database was held on the server and kodi lazily loaded this data. If kodi caches the data locally then even better! Although I would like to aim for this to be more of a __background__ activity and updates populate the kodi cache/UI much more readily.

Ill update with your suggestions of comments.
On maintainability: Ill leave the location and implementation details up to yourself; whatever is easiest the structure of the project.

mcarlton00:
If you don't want to appear condescending, the tone of your writing could use an adjustment.  Bolding words and passive-aggressive comments aren't winning you any favors. What's obvious to you may not be obvious to me or any of the other Jellyfin contributors.  We're a community of varied backgrounds, strengths, and skill levels.  Just be mindful of others, that's all.  I don't ask questions to be annoying or hard to work with. I ask questions to make sure I understand the changes and that they make sense for the project as a whole.  As you may have noticed from being in the code the last few days, it's not always laid out in a super straightforward way and there's some places that don't make any sense at all.  We make improvements as we can, but we're all volunteers with limited time, it's a fairly large undertaking, and everybody currently involved is relatively new to the codebase ourselves, so it's a constant learning experience for everyone.

With all that said and out of the way, let's put it rest and move on to more important things. We're generally always going to be in favor of performance increases, and this one is no exception.  From my understanding, it seems as though this could possibly be solved as a function in `utils.py`.  Something like
```python
def pprint_json(params):
    return json.dumps(params, indent=4)
```
I'm not afraid to admit when I don't know something, and you're obviously more experienced with low level python and performance than I am.  That's why I'm asking you if a function would solve the same problem, or if it has to be a class?  Either way, I would like to see it moved to one of the existing files, likely the previously mentioned `utils.py` unless you can think of a good reason it shouldn't.  This isn't a dictatorship, I'm (and the rest of the Jellyfin team) fully willing to discuss pros and cons of various decisions.  From my point of view, adding another file to the repo for essentially 6 lines of code just makes one more piece that we have to keep in mind and adds to the complexity.  A small addition, sure, but that's likely how the codebase got to it's current form.

----------------------------
As some general background and knowledge that may or may not help you moving forward:
* Perhaps obvious, Jellyfin runs it's own database on the server containing all metadata for your media
* The JF for Kodi addon does maintain it's own local database, as you've obviously noticed.  However because we're modifying the Kodi database directly it's not allowed into the official addon repos, hence us running our own.
* There was another addon back in the Emby days that would load content directly from the server instead of syncing it locally, but so far there hasn't been a concentrated interest to fork and maintain that for Jellyfin.  The performance of having metadata synced locally is infinitely better than lazy loading it from the server, as you can imagine.  But if you're interested, that addon can be found [here](https://github.com/faush01/plugin.video.embycon)
* Since you mentioned library syncing/updates:
  * When Kodi is running and the Jellyfin server receives updates of any kind, those changes are synced to the Kodi addon by communicating over websockets.  This means that watched status and new media should appear in a local Kodi install fairly quickly after it's changed in the server.
  * When new media/watched status changes occur while Kodi _isn't_ running, it relies on the endpoints exposed by the [KodiSyncQueue server plugin](https://github.com/jellyfin/jellyfin-plugin-kodisyncqueue) to determine what information needs to be updated

This ended up being way longer than expected.  If you're still reading, then cheers, happy bug hunting, and feel free to join us in Matrix chat to discuss things in more "real time" if you'd like

druscoe:
Lol. I didn't intend the tone my messages to come across as you have read them. The formatting is purely for emphasis. I'm sorry if any offence was made.

Unfortunately, the pprint_json will not work as the dumps is still being called at too early. Imagine the logging is something like the following pseudocode:

```
1:  def log_info_message(formatting, message):
2:      if logging_level == INFO:
3:          print(formatting % message)
4:
5:  log_info_message("%s", str(True))
6:  log_info_message("%s", True)
```

On line 5 the value is explicitly converted to the string 'true' and passed into the function as a string.
On line 6 the value True is not formatted into a string and is passed into the function as a boolean. The conversion to a string is deferred until line 3. This conversion does not happen unless the logging info is DEBUG.

This is a trivial example in which converting a bool to a string is nominal. For a larger conversion to a string, we really want to emulate the behaviour of the call on line 6. We want to pass an object which is potentially convertible to string and only evaluate the string representation if it is needed.

This is what the _ _ str _ _ function of an object does. A string is returned when a string representation of that object is requested from the object. JsonDebugOutput is an implementation of the Adapter design pattern; a wrapper to glue a call from str(...) to a call to json.dumps(...). I cannot think of another way of achieving this deferred behaviour without a wrapper class as it is implemented.

mcarlton00:
You could've just said "No, it needs a class" and I would've trusted you, but I appreciate the in depth explanation :laughing: I'm on the same page now.

I think if we move the `JsonDebugPrinter` to `utils.py` then we'll be good to merge this one.  Mostly because it's already there and I don't see the benefit of adding another file with so little content.  I could be convinced otherwise, but that's my preference at this moment.  And a short comment about why it exists, please.

druscoe:
Ive moved the class into the existing utils module.
The results of this should be more significant now that other optimizations have been applied. To see the effect of deferring the json conversion try importing your library with logging level debug and logging level info. Could you also share the difference you see in time with between debug and info levels.
Thanks

mcarlton00:
Perfect.  Tests again done using a raspberry pi 3b.  I'm seeing about 45 seconds difference between info and debug modes across that ~1000 item movies library.  Which if I'm doing my math correctly and the difference stays linear, extends out to roughly 15 minutes for syncing my entire video library.

druscoe:
I was hoping for more noticeable differences than the couple of seconds reported previously. Although you are getting better results my testing is showing much more significant increases. Maybe we are using different Json libraries? Ive found requests module uses simplejson module on my dev machine. I'll investigate which library I get when I import json directly.

Thanks for your results

mcarlton00:
Those results were purely with this PR, not all of your recent changes together.  Now that it's in master and I've tested all of them together.

0.4.0 (current release) - 6:22 sync time
Current master (debug logging) - 5:45 sync time
Current master (info logging) - 4:40 sync time

It's a bit odd that I'm getting slower results with everything merged than I was with just #200, but it's still _significant_ improvement over the previous sync times

druscoe:
> Those results were purely with this PR, not all of your recent changes together. Now that it's in master and I've tested all of them together.
> 
> 0.4.0 (current release) - 6:22 sync time
> Current master (debug logging) - 5:45 sync time
> Current master (info logging) - 4:40 sync time
> 
> It's a bit odd that I'm getting slower results with everything merged than I was with just #200, but it's still _significant_ improvement over the previous sync times

Because it takes so long I am inclined not to go and get the actual metrics. From my memory though, my testing has brought a full sync down from ~1h30m to about 7m on my android box. It is odd that I am seeing such huge differences when compared to your results?

I may have an explanation for the 'increase' in time since #200.
I noticed #201 is adding ~30~40s on my dev machine as sqlite3.cursor.close() is being called many more times. This difference may be increased to minutes on a RPi. I have been playing with the idea of trying to only lock/unlock the kodi database as that is the only one which needs read access to update the DB -- This will be safe as if we havent got the lock then we are not writing to it.
The other option is to revert #201. A viable option as the feature is not fully implemented.

Any thoughts?
Thanks

mcarlton00:
There's still several things that need to be fixed before I'm ready to push another release, so we're not quite under a time crunch.  TrueTechy mentioned finding another possible bug yesterday while testing something else that we need to look into, for example.  And there is significantly more speed now than in the current release, so I'm not opposed to leaving 201 in regardless.  Personally, I think if we want the UI to update during syncing it's going to increase the sync time regardless of how it's implemented and we'll just have to accept that.

ps: Mentioning again to feel free to join us in the [Matrix Chat](https://matrix.to/#/#jellyfin:matrix.org), as it's a bit more convenient for more conversational topics

PR #200, opened 2020-02-18
Resort to the expensive database lookup only if the person exists in the

druscoe:
database.

The database is configured such that each person name within the table
has to be unique. Attempting to add a person which already exists causes
an IntegrityError exception to be thrown. This is a less costly
operation than the cursor.execute and cursor.fetch operations.

Increase performance by reverting to the expensive calls of fetching if
adding a person fails.

This pattern could also be applied to other tables within the database.


jellyfin_kodi/objects/kodi/kodi.py
```diff
@@ -21,6 +22,14 @@ class Kodi(object):
     def __init__(self):
         self.artwork = artwork.Artwork(self.cursor)
 
+        try:
+            self.cursor.execute(QU.get_all_people)
+        except:
+            # Failed to load the table. Has the table been created?
+            self._people_cache = {}
+        else:
+            self._people_cache = dict(self.cursor.fetchall())
```
mcarlton00: Wouldn't it make more sense to move this up above the `except`?  I've never found a need for the else block over just putting the desired actions in the try block, so I'm curious about the choice to include it here.

jellyfin_kodi/objects/kodi/kodi.py
```diff
@@ -123,46 +132,59 @@ def add_thumbnail(person_id, person, person_type):
...
         return person_id
 
-    def get_person(self, *args):
-
+    def _get_person(self, *args):
         try:
+            return self.add_person(*args)
+        except IntegrityError:
+            # The person already exists in the database
+            # Now we can do the expensive operation of fetching the id
```
mcarlton00: Shouldn't this block also be checking `_people_cache` first, and then doing the add if they don't already exist?  Similar to the `get_person` below

jellyfin_kodi/objects/kodi/kodi.py
druscoe: I believe it is a good pattern to use the else block of a try/catch.
It helps communicate to a reader which calls the author intended to catch from and which calls they did not anticipate catching from. It also prevents potential accidental catches of exceptions from calls you may want to propagate or learn about.
I think I read somewhere to keep the try blocks as small as possible but no smaller; Ive taken that to heart as I have (albeit small) justifications for using this pattern but not seen evidence against using it.

jellyfin_kodi/objects/kodi/kodi.py
druscoe: _get_person is internal/private. Ideally, it's an implementation detail which ideally should only be called by get_person.
If another callee calls this directly then I can only assume that they do not want to access the cache. This is the only case where the catch block should get called as calling _get_person via get_person means that the person does not exist in the db and add_person should'nt raise an IntegrityError

druscoe:
> Very dramatic speed increase. From 6:20 down to 3 minutes flat for my movie library on a Pi 3B. Nicely done. Couple questions, but otherwise it looks good.
Thanks

> I don't really see a whole lot of benefit doing this for most of the other tables, since obviously movies/tv shows are more "unique" versus actors that can be in many things. But the other one that stands out that could have another speed boost would be studios. Possibly music artists as well? (Food for thought for another PR later)
I like consistency but maybe best not to force this pattern on other tables if they don't need it. 

Artists already get put in the people table (I believe) as I have noticed an improvement on adding albums since this change -- Disclosure; This is a presumption from library updates and not from reading code nor profiling.

Studios for sure! genres, bands*, albums**, songs are the tables I think may benefit.
* Are bands stored in another table, separate from people?
**Maybe
These tables need profiling first before any action.
Any others which you may think of? -- I was guided to the people table by the profiler so that may help identify others.

jellyfin_kodi/objects/kodi/kodi.py
mcarlton00: I see what you mean now.  I should probably look at the full file instead of just the little individual changed snippets

jellyfin_kodi/objects/kodi/kodi.py
mcarlton00: Doing some quick research, it does appear that you're correct and using the else block is considered more "proper".  I've just never seen it done this way before.

mcarlton00:
I've found it easier to check the database directly for these types of things rather than the code.
```
kronos% sqlite3 .kodi/userdata/Database/MyVideos116.db
SQLite version 3.31.1 2020-01-27 19:55:54
Enter ".help" for usage hints.
sqlite> .tables
actor                   movie_view              studio_link
actor_link              movielinktvshow         tag
art                     musicvideo              tag_link
bookmark                musicvideo_view         tvshow
country                 path                    tvshow_view
country_link            rating                  tvshowcounts
director_link           season_view             tvshowlinkpath
episode                 seasons                 tvshowlinkpath_minview
episode_view            sets                    uniqueid
files                   settings                version
genre                   stacktimes              writer_link
genre_link              streamdetails
movie                   studio
```
I created #203 to track this further.  That will allow for more overarching discussion rather than individual PRs.

PR #201, opened 2020-02-18
Allow for UI updates as the library is being sync'ed

druscoe:
Periodically release the locks on the database as the sync with the
server is being performed. This allows the UI to communicate with its
database and update the UI in real time.

Taken this oppurtunity to refactor some common code improve
maintainance.

### DIFF
--- a/jellyfin_kodi/full_sync.py
+++ b/jellyfin_kodi/full_sync.py
@@ -3,6 +3,7 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 
 ##################################################################################################
 
+from contextlib import contextmanager
 import datetime
 import logging
 
@@ -248,6 +249,13 @@ class FullSync(object):
 
             raise
 
+    @contextmanager
+    def video_database_locks(self):
+        with self.library.database_lock:
+            with Database() as videodb:
+                with Database('jellyfin') as jellyfindb:
+                    yield videodb, jellyfindb
+
     @progress()
     def movies(self, library, dialog):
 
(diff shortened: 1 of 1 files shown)


mcarlton00:
Perhaps I'm missing something, but I'm not seeing much change from the UI side.  I still have the library update counter spinning in the corner during the sync, but I'm not seeing any change on the home screen or in the libraries until the sync is finished.  Where exactly is this allowing the UI to update during a sync?

druscoe:
Admittedly this change is not working as I had originally hoped for but there are still "some" improvements:
- As a library is updating (with an initial empty kodi library) the user browsed can browse into Movies and the list will remain empty until the sync has completed. With this patch the user will see the movies which have been sync'ed thus far. The same is true as from a non-empty initial library.
- Browsing the "Jellyfin dynamic" hung while the sync was working (sometimes even crashing if I tried to cancel). Access to the Jellyfin dynamic now works as expected. -- Full disclosure here, I've seen this a few times but not concentrated any serious testing on it; just something I happened to notice.
- Refactored the code to reduce duplication.

Its not the full feature set which I was hoping for but enough improvements IMO to be committed as an intermediary. This patch will complement my planned future patch when I can find the code which updates Kodi's UI. The complete idea is that you will see new movies appearing on your screen as the sync is happening and without any user interaction.

mcarlton00:
I'm pretty sure what you're looking for is `xbmc.executebuiltin('ReloadSkin()')` to reload the home screen.  I'm not sure where the best place to implement it here would be, but that should point you in the right direction.  Only problem I see with using that is that it fully reloads the screen, so it automatically jumps the cursor back up to the top entry in the list, which could be annoying for the user.

druscoe:
> I'm pretty sure what you're looking for is `xbmc.executebuiltin('ReloadSkin()')` to reload the home screen. I'm not sure where the best place to implement it here would be, but that should point you in the right direction. Only problem I see with using that is that it fully reloads the screen, so it automatically jumps the cursor back up to the top entry in the list, which could be annoying for the user.
> 
> Full list of supported builtin commands are [here](https://kodi.wiki/view/List_of_built-in_functions)

Indeed. I wanted to test and see what the refresh is like. Im a bit worried that if I could get the UI to display updated media files without moving the current selection that the visual updates will be too distracting on their own, especially if it is happening at an increased pace now.

The effect I am aiming for is just a refresh of missing data as opposed to a full refresh. There must be a way of informing Kodi that the database has been updated without calling a full ReloadSkin? If it exists then I'll discover it :) Maybe it requires inbuilt skin support?

mcarlton00:
I think the issue you're going to run into is that the home screen can't update in real time (afaik).  It's kinda an all or nothing deal.  For example, with this PR as it stands as things are syncing I can move into the movies directory, then back to the home screen, and there's media displayed there.  It needs a reload of that screen of some kind before new media can show up.

Maybe it could make sense to trigger a reload every ~100 items synced or some arbitrary number we choose so periodic updates do occur, but it's not a constant annoying flashing.  That approach does need more testing first though.  Specifically:
* Try to keep the user's location on the menu (or maybe catch it's location and move back there after the reload finishes)
* What happens when a skin is reloaded when inside a library view?
* What happens when a skin is reloaded while video playback is happening?

I guess at this point the options are either mark this PR as a draft and keep working on it, or open an issue to keep track of the new ideas and we'll merge this one

druscoe:
> * Try to keep the user's location on the menu (or maybe catch it's location and move back there after the reload finishes)
> * What happens when a skin is reloaded when inside a library view?
> * What happens when a skin is reloaded while video playback is happening?

I know what an annoyance constant and interfering refreshes can be; they can bring the UI to a point of unusability. So, unless I can get all these points listed here then I will not implement it. Maybe a call to ReloadSkin and rebuild the users' selection etc? This will be hard to get right though.
 
> I guess at this point the options are either mark this PR as a draft and keep working on it, or open an issue to keep track of the new ideas and we'll merge this one

It's good enough in that the user can see recent updates, albeit manually. For that reason, I'm tending towards the latter option of merging and opening an issue.

Issue #206, opened 2020-02-23
Kodi 19 - Unable to sync libraries


mcarlton00:
**Describe the bug**

Running the addon in Kodi 19 is failing to sync libraries.
**To Reproduce**

1. Authenticate to server
2. Select libraries to sync
3. Click OK
4. No progress

**Logs**

```
2020-02-23 03:13:08.206 T:1740638  NOTICE: JELLYFIN.full_sync -> starting sync with ['f137a2dd21bbc1b99aa5c0f6bf02a805', '767bffe4f11c93ef34b805451a696a4e', '7e64e319657a9516ec78490da03edccb', 'Boxsets:']
2020-02-23 03:13:08.219 T:1740638  NOTICE: JELLYFIN.helper.wrapper -> Processing Movies: f137a2dd21bbc1b99aa5c0f6bf02a805
2020-02-23 03:13:08.262 T:1740638  NOTICE: JELLYFIN.full_sync -> name 'xrange' is not defined
                                            Traceback (most recent call last):
                                              File "jellyfin_kodi/full_sync.py", line 233, in process_library
                                                media[library['CollectionType']](library)
                                              File "jellyfin_kodi/helper/wrapper.py", line 41, in wrapper
                                                result = func(self, dialog=dialog, *args, **kwargs)
                                              File "jellyfin_kodi/full_sync.py", line 266, in movies
                                                for items in server.get_items(library['Id'], "Movie", False, self.sync['RestorePoint'].get('params')):
                                              File "jellyfin_kodi/downloader.py", line 182, in get_items
                                                for items in _get_items(query):
                                              File "jellyfin_kodi/downloader.py", line 270, in _get_items
                                                for offset in xrange(params['StartIndex'], items['TotalRecordCount'], LIMIT)]
                                            NameError: name 'xrange' is not defined
2020-02-23 03:13:08.265 T:1740638  NOTICE: JELLYFIN.full_sync -> --<[ fullsync ]
2020-02-23 03:13:08.266 T:1740638  NOTICE: JELLYFIN.library -> argument of type 'NameError' is not iterable
                                            Traceback (most recent call last):
                                              File "jellyfin_kodi/full_sync.py", line 233, in process_library
                                                media[library['CollectionType']](library)
                                              File "jellyfin_kodi/helper/wrapper.py", line 41, in wrapper
                                                result = func(self, dialog=dialog, *args, **kwargs)
[...]

PR #207, opened 2020-02-23
Make commands python 2/3 compatible

mcarlton00:
Fixes syncing issues mentioned in #206

### DIFF
--- a/jellyfin_kodi/downloader.py
+++ b/jellyfin_kodi/downloader.py
@@ -15,6 +15,14 @@ from jellyfin import Jellyfin
 from jellyfin import api
 from jellyfin.exceptions import HTTPException
 
+# Python 2/3 Compatible things
+import sys
+if sys.version.startswith('3'):
+    xrange = range
+    izip = zip
+else:
+    from itertools import izip
+
 #################################################################################################
 
 LOG = logging.getLogger("JELLYFIN." + __name__)
@@ -269,7 +277,6 @@ def _get_items(query, server_id=None):
         query_params = [get_query_params(params, offset, LIMIT) \
                 for offset in xrange(params['StartIndex'], items['TotalRecordCount'], LIMIT)]
 
-        from itertools import izip
         # multiprocessing.dummy.Pool completes all requests in multiple threads but has to 
         # complete all tasks before allowing any results to be processed. ThreadPoolExecutor
         # allows for completed tasks to be processed while other tasks are completed on other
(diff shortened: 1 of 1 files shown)


druscoe:
I have much more experience in python2 opposed to python3 so this is my time to learn :)
Is xrange and izip no longer supported in python3?

A concern I have looking at this change is 
Do moves.zip and moves.range return iterators or lists?

Thinking in terms of python2 zip and izip
The usage of izip was important here as it allowed the addon to be processing data while the thread pool was still working in the background. izip returns an iterator which can be iterated before the entire sequence is built.
If this was changed to zip then all threads would need to complete for the result to be built into a list. Processing on the data will only occur after all threads have been executed.

The xrange is much less important

TrueTechy:
xrange became range in python3 as it is just much better than the range in python2. Apparently a similar thing happened with zip and izip

https://stackoverflow.com/questions/32659552/importing-izip-from-itertools-module-gives-nameerror-in-python-3-x

druscoe:
Issue #206 
Works fine on my machine :) lol
Sorry for the breakage.

I was under the assumption that python was shipped with kodi and that version was python3.
I am using Kodi 17.6 on my dev machine and Kodi 18.5 on my android box; neither complain about xrange and izip

My question is how can I guarantee that my patches work with all version of Kodi shipping with different versions of python? Creating bugs is just embarrassing :/

druscoe:
> xrange became range in python3 as it is just much better than the range in python2. Apparently a similar thing happened with zip and izip
> 
> https://stackoverflow.com/questions/32659552/importing-izip-from-itertools-module-gives-nameerror-in-python-3-x

Thank you for this info.

A further clarification if you dont mind.
moves is a util library which helps migrate from python2 to python3. So on a python3 platform moves.zip will be an "alias" to python3 zip and hence return an iterator as opposed to a list.

On a python2 platform is zip "aliasing" itertools.izip or builtin.zip?

mcarlton00:
It's not a problem, I just forgot to check the other day and was starting to look at what needed to be wrapped up for our next release and noticed it.  Python 3 in Kodi land is a relatively new development, currently only available on Kodi 19.  Which depending on your system is either an easy download or a manual compilation process.

Python 3 shuffled lots of stuff around.  Mostly for the better, but it does make supporting a codebase on both versions mildly annoying.  That's where the six library comes into play and helps with that greatly, otherwise you do silly things like my first commit on this PR (forgot six existed for a while there).  You can see more details about `six.moves` at [this section](https://six.readthedocs.io/#module-six.moves) of their docs.  Specifically if you scroll down there's a big table with the python 2/python 3 equivalents.  And yes, it's using `itertools.zip` in python 2 (second from the bottom entry in the list)

PR #208, opened 2020-02-23
Simplify request interface

druscoe:
Hi,

During optimization investigations, I was coming to the point where the network was my biggest bottleneck. Attempting to deduce where could be a chance for optimization was daunting as the interface to http.request was cumbersome. The author had decided to pass in a chunk of data in a dictionary for what appears to be a design decision for extensibility and abstraction for many different types of calls (ie. login, authentication and general server queries)
I was unable to reason if parameters were purely "in" parameters or "in/out" parameters and such was bound to the data structures that were already in place.

This PR is only a refactorization (sorry). The main aim to use pythons inbuilt parameter passing mechanism instead of an ad-hoc dictionary. It will enable me to investigate potential optimizations in the http/networking code for general server queries while reducing the risk that authentication (say) will be broken.

### DIFF
--- a/jellyfin_kodi/jellyfin/connection_manager.py
+++ b/jellyfin_kodi/jellyfin/connection_manager.py
@@ -99,16 +99,13 @@ class ConnectionManager(object):
             raise AttributeError("server cannot be empty")
 
         try:
-            request = {
-                'type': "POST",
-                'url': self.get_jellyfin_url(server, "Users/AuthenticateByName"),
-                'json': {
-                    'Username': username,
-                    'Pw': password or ""
-                }
+            url = self.get_jellyfin_url(server, "Users/AuthenticateByName")
+            json = {
+                'Username': username,
+                'Pw': password or ""
             }
+            result = self._REQUEST_URL(url, "POST", json=json, additional_headers=False)
 
-            result = self._request_url(request, False)
         except Exception as error:  # Failed to login
             LOG.exception(error)
             return False
@@ -155,6 +152,7 @@ class ConnectionManager(object):
             return self._after_connect_validated(server, credentials, result, True, options)
 
         except Exception as e:
+            raise
             LOG.info("Failing server connection. ERROR msg: {}".format(e))
             return { 'State': CONNECTION_STATE['Unavailable'] }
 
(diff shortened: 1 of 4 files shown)


TrueTechy:
I agree that the networking in the addon needs a lot of cleaning up. I'd started on a PR to refactor the connection manager so that all of the calls to the server are actual in the API.py file. Would help when the big API change comes on the server side as well

mcarlton00:
Refactoring has been the name of the game in a lot of the big PRs that have been merged so far.  Manageable code is better code.

Nothing is standing out to me as an issue, but @TrueTechy has spent a lot more time than me in this part of the code than I have and was in progress of refactoring other pieces of it, so I'll defer to them for approving this one.

jellyfin_kodi/jellyfin/http.py
```diff
@@ -49,48 +49,60 @@ def stop_session(self):
...
-        '''
-        if not data:
-            raise AttributeError("Request cannot be empty")
-
-        data = self._request(data)
-        LOG.debug("--->[ http ] %s", JsonDebugPrinter(data))
-        retry = data.pop('retry', 5)
+
+    def get_handler_url(self, handler):
+        server = self.config.data.get("auth.server", "")
```
TrueTechy: Looks like the rest of the instances of this have had the self and .data removed. Should this one not also have that?

TrueTechy:
@druscoe You still about to comment on this?

jellyfin_kodi/jellyfin/connection_manager.py
```diff
@@ -243,18 +217,17 @@ def _connect_to_servers(self, servers, options):
...
-            'type': "GET",
-            'url': url,
-            'dataType': "json",
-            'timeout': timeout,
-            'verify': options.get('ssl'),
-            'retry': False
-        })
+        try:
+            return self.http.request_url(url, "GET", headers=headers, \
+                    timeout=timeout, retry=False)
```
oddstr13: Newline is allowed between function parameters, there is no need to escape the newline.

jellyfin_kodi/jellyfin/connection_manager.py
```diff
@@ -423,15 +396,13 @@ def _after_connect_validated(self, server, credentials, system_info, verify_auth
...
-                'headers': {
-                    'X-MediaBrowser-Token': server['AccessToken']
-                }
-            })
+            url = self.get_jellyfin_url(server['address'], "System/Info")
+            headers = self._get_headers()
+            headers['X-MediaBrowser-Token'] = server['AccessToken']
+
+            system_info = self.http.request_url(url, "GET", headers=headers, \
+                    verify=options.get('ssl'), timeout=self.timeout)
```
oddstr13: Newline escape.

jellyfin_kodi/jellyfin/http.py
```diff
@@ -49,48 +49,60 @@ def stop_session(self):
...
+        return "%s/%s" % (server, handler)
+
+
+    def request_url(self, url, type, params=None, json=None, session=None, \
+                    headers={}, verify=None, timeout=None, retry=5):
```
oddstr13: Newline escape.

jellyfin_kodi/jellyfin/http.py
oddstr13: `_replace_user_info` is getting `config` as a parameter, `get_handler_url` is not.

PR #209, opened 2020-02-23
Reduce execute and fetch one calls

druscoe:
Hi,

Im opening this PR for input from other devs if that is okay.

Ive tried to reduce the number of calls to sqlite3.execute and sqlite3.fetchone as profiling has highlighted these two as the biggest bottlenecks. I am managing to reduce the number of calls and delete code :) but the profiling is still giving consistent results on the amount of time spent executing each function. Is the effort of continuing with idea worth the effort?

Please see the commit logs for both commits as they contain performance metrics.

Thanks
Damo

### DIFF
--- a/jellyfin_kodi/objects/kodi/kodi.py
+++ b/jellyfin_kodi/objects/kodi/kodi.py
@@ -40,11 +40,6 @@ class Kodi(object):
 
         return self.cursor.fetchone()[0] + 1
 
-    def create_entry_person(self):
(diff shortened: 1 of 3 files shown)


TrueTechy:
The addon writes to both the jellyfin database and the kodi database for every item in both. I wonder if we could find where the specific code that's bottleneck it is given the reduction thus far is in the 100s of millisecond ranges it might be wise to spend time working on issues elsewhere?

Issue #153, opened 2019-12-07
Date errors during sync


develroo:
When syncing TV shows this error sometimes appears, though no corresponding error happens on the server side. 

**To Reproduce**

Just begin fresh sync.
Watch the logs.

**Expected behavior**
Error free logs (Oh wouldn't that be nice? )

**Logs**
 ```
2019-12-07 21:39:10.644 T:140124612318976  NOTICE: JELLYFIN.helper.utils -> ERROR::resources/lib/helper/utils.py:484 date value out of range
                                            Traceback (most recent call last):
                                              File "resources/lib/helper/utils.py", line 480, in convert_to_local
                                                date = date.astimezone(tz.tzlocal())
                                              File "/home/user/.var/app/tv.kodi.Kodi/data/addons/script.module.dateutil/lib/dateutil/tz/_common.py", line 144, in fromutc
                                                return f(self, dt)
                                              File "/home/user/.var/app/tv.kodi.Kodi/data/addons/script.module.dateutil/lib/dateutil/tz/_common.py", line 258, in fromutc
                                                dt_wall = self._fromutc(dt)
                                              File "/home/user/.var/app/tv.kodi.Kodi/data/addons/script.module.dateutil/lib/dateutil/tz/_common.py", line 222, in _fromutc
                                                dtoff = dt.utcoffset()
                                              File "/home/user/.var/app/tv.kodi.Kodi/data/addons/script.module.dateutil/lib/dateutil/tz/tz.py", line 216, in utcoffset
                                                if self._isdst(dt):
                                              File "/home/user/.var/app/tv.kodi.Kodi/data/addons/script.module.dateutil/lib/dateutil/tz/tz.py", line 288, in _isdst
                                                if self.is_ambiguous(dt):
                                              File "/home/user/.var/app/tv.kodi.Kodi/data/addons/script.module.dateutil/lib/dateutil/tz/tz.py", line 250, in is_ambiguous
                                                (naive_dst != self._naive_is_dst(dt - self._dst_saved)))
                                            OverflowError: date value out of range
     2019-12-07 21:39:11.368 T:140124612318976  NOTICE: Previous line repeats 3 times.
````

**System (please complete the following information):**
 - OS: Debian
 - Jellyfin Version: 10.4.1
 - Kodi Version: 18.4 Git:20190906-nogitfound
 - Addon Version: 0.3.6

**Comment by mcarlton00** [comment]
Along the same lines, I've seen a few of these show up.
```
[...]

PR #212, opened 2020-02-26
Bad dates are bad

mcarlton00:
When an item from the server doesn't have a proper date, it seems to get set as something along the lines of `'PremiereDate': u'0001-01-01T00:00:00.0000000Z'`, which obviously isn't valid and was throwing errors.  It's harmless, but annoying.  This makes it do a check to see if the year is set to this default value and prevents throwing errors on it.

Also adds the date to the error message in the hopes of tracking down the cause of #153

### DIFF
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -478,13 +478,13 @@ def convert_to_local(date):
     '''
     try:
         date = parser.parse(date) if isinstance(date, string_types) else date
-        parsed_date = date.replace(tzinfo=tz.tzutc())
-        local_date = parsed_date.astimezone(tz.tzlocal())
+        date = date.replace(tzinfo=tz.tzutc())
+        date = date.astimezone(tz.tzlocal())
         # Bad metadata defaults to date 1-1-1.  Catch it and don't throw errors
-        if local_date.year == 1:
+        if date.year == 1:
             return str(date)
         else:
-            return local_date.strftime('%Y-%m-%dT%H:%M:%S')
+            return date.strftime('%Y-%m-%dT%H:%M:%S')
     except Exception as error:
         LOG.exception('Item date: {} --- {}'.format(str(date), error))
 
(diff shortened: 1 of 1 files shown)


flatline-studios:
I've just come across a similar issue, in that some of my videos (home movies), had erroneous year tags (it had put together the whole date in the year field in the metadata).

My Kodi log would show:

`"ProductionYear": 20110112, `

when run with debug logging in the jellyfin settings.

It would then produce this error:

```
2020-02-27 16:39:00.274 T:20092  NOTICE: JELLYFIN.full_sync -> ERROR::jellyfin_kodi\full_sync.py:240 year is out of range
                                              File "jellyfin_kodi\full_sync.py", line 231, in process_library
                                              File "jellyfin_kodi\helper\wrapper.py", line 40, in wrapper
                                              File "jellyfin_kodi\full_sync.py", line 363, in musicvideos
                                                obj.musicvideo(mvideo, library=library)
                                              File "jellyfin_kodi\helper\wrapper.py", line 102, in wrapper
                                                return func(*args, **kwargs)
                                              File "jellyfin_kodi\helper\wrapper.py", line 116, in wrapper
                                              File "jellyfin_kodi\helper\wrapper.py", line 175, in wrapper
                                                return func(self, item, *args, **kwargs)
                                              File "jellyfin_kodi\objects\musicvideos.py", line 79, in musicvideo
                                                obj['Premiere'] = Local(obj['Premiere']) if obj['Premiere'] else datetime.date(obj['Year'] or 2021, 1, 1)
                                            ValueError: year is out of range
```

Does this commit fix that error as well? Or only the Premiere Date?

mcarlton00:
This commit only fixed things within the `convert_to_local` function, which I don't see in that traceback.  So unfortunately it likely won't fix it.  Also confusing that it's reporting as a music video (unless that's the kind of library you used in Jellyfin).  Either way, this is likely worth opening a new issue with those details so it can be tracked independently

flatline-studios:
It was listed under music video in Jellyfin, as there isn't any generic/no metadata needed/whatever category.

PR #232, opened 2020-03-15
Mitigation for #216 for invalid year

### LINKED ISSUE #216: Bad year metadata causing sync failure
Some of my videos (home movies), had erroneous year tags (it had put together the whole date in the year field in the metadata).

My Kodi log (with Jellyfin addon debug enabled) shows:

`"ProductionYear": 20110112, `

It then produces this error:

```
2020-02-27 16:39:00.274 T:20092  
NOTICE: JELLYFIN.full_sync -> ERROR::jellyfin_kodi\full_sync.py:240 year is out of range
    File "jellyfin_kodi\full_sync.py", line 231, in process_library
    File "jellyfin_kodi\helper\wrapper.py", line 40, in wrapper
    File "jellyfin_kodi\full_sync.py", line 363, in musicvideos
      obj.musicvideo(mvideo, library=library)
    File "jellyfin_kodi\helper\wrapper.py", line 102, in wrapper
      return func(*args, **kwargs)
    File "jellyfin_kodi\helper\wrapper.py", line 116, in wrapper
    File "jellyfin_kodi\helper\wrapper.py", line 175, in wrapper
      return func(self, item, *args, **kwargs)
    File "jellyfin_kodi\objects\musicvideos.py", line 79, in musicvideo
      obj['Premiere'] = Local(obj['Premiere']) if obj['Premiere'] else datetime.date(obj['Year'] or 2021, 1, 1)
  ValueError: year is out of range
```

Can we get a sanity check for the value(s) of the date/year fields, and simply ignore the bad data if it fails?

TrueTechy:
Adds a mitigation which will close #216 where the sync fails for media when the year data received from the server is not a valid year as the server doesn't valid this data at this time

### DIFF
--- a/jellyfin_kodi/downloader.py
+++ b/jellyfin_kodi/downloader.py
@@ -5,6 +5,7 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 
 import logging
 import threading
+import concurrent.futures
 
 from six.moves import range, queue as Queue, zip
 
@@ -273,7 +274,6 @@ def _get_items(query, server_id=None):
         # complete all tasks before allowing any results to be processed. ThreadPoolExecutor
         # allows for completed tasks to be processed while other tasks are completed on other
         # threads. Dont be a dummy.Pool, be a ThreadPoolExecutor
-        import concurrent.futures
         p = concurrent.futures.ThreadPoolExecutor(DTHREADS)
 
         results = p.map(lambda params: _get(url, params, server_id=server_id), query_params)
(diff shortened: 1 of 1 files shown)


jellyfin_kodi/downloader.py
```diff
@@ -282,6 +282,17 @@ def get_query_params(params, start, count):
             query['params'] = params
 
             result = result or {'Items': []}
+
+            # Mitigates #216 till the server validates the date provided is valid
+            for count in range(0, len(result["Items"])):
```
mcarlton00: With the threading above, this shouldn't need to be a loop.  It will always only be processing 1 item at a time thanks to the concurrent library.

jellyfin_kodi/downloader.py
```diff
@@ -282,6 +282,17 @@ def get_query_params(params, start, count):
...
+                if result["Items"][count].get('ProductionYear'):
```
mcarlton00: Mixing single and double quotes.  tsk tsk

I prefer single in most cases, though the addon is already mixing them all over the place.  We just need to decide on one and be consistent with it.

jellyfin_kodi/downloader.py
```diff
@@ -282,6 +282,17 @@ def get_query_params(params, start, count):
...
+                    try:
+                        from datetime import date
```
mcarlton00: Move the import to the top of the file.  Here it's being re-imported over and over again as each item is processed

jellyfin_kodi/downloader.py
oddstr13: Personally, I tend to use single quote for dictionary indexing, and double quote for "normal" strings. No particular reason why, it's just how I've been doing it over the years..

PR #233, opened 2020-03-18
Updated address normalization to test for https first to fix #223

### LINKED ISSUE #223: HTTPs redirection issue when logging in
**Describe the bug**
When using a reverse proxy to redirect from http to https the user can't login with the server returning a 500 error, if the https is not specified by the user in the URL

**To Reproduce**
1. Attempt to login

**Expected behavior**
Successful Login

**Logs**
```
2020-03-06 23:56:59.516 T:139967855077120  NOTICE: JELLYFIN.jellyfin.api -> ERROR::jellyfin_kodi/jellyfin/api.py:425 Failed to login to server with status code: 500
2020-03-06 23:56:59.517 T:139967855077120  NOTICE: JELLYFIN.jellyfin.api -> ERROR::jellyfin_kodi/jellyfin/api.py:426 Server Response:
                                            Input string was not in a correct format.
```
 - OS: Ubuntu
 - Jellyfin Version: 10.4.3
 - Kodi Version: 18.5
 - Addon Version: development branch

TrueTechy:
Resolves #223 where the http->https redirect was causing the server to not respond properly. This changes the default to try HTTPS first, then if that is not available goes down to http

### DIFF
--- a/jellyfin_kodi/jellyfin/connection_manager.py
+++ b/jellyfin_kodi/jellyfin/connection_manager.py
@@ -134,6 +134,10 @@ class ConnectionManager(object):
 
         try:
             public_info = self.API.get_public_info(address)
+            if not public_info:
+                LOG.error("connectToAddress %s failed", address)
+                return { 'State': CONNECTION_STATE['Unavailable'] }
+
             LOG.info("connectToAddress %s succeeded", address)
             server = {
                 'address': address,
@@ -286,18 +290,18 @@ class ConnectionManager(object):
         # Attempt to correct bad input
         url = urllib3.util.parse_url(address.strip())
 
-        # Default to using https
+         # Default to using https
         url = url._replace(scheme='https')
-
+      
         # Test if server is reachable over https
         if self.API.get_public_info(url):
-            if url.scheme == 'https' and url.port == 443:
+            if url.port == 443:
                 url = url._replace(port=None)
         else:
             # Server didn't give an expected response over https
             # Try http instead
-            url._replace(scheme='http')
-            if url.scheme == 'http' and url.port == 80:
+            url = url._replace(scheme='http')
+            if url.port == 80:
                 url = url._replace(port=None)
 
         return url.url
--- a/jellyfin_kodi/jellyfin/connection_manager.py
+++ b/jellyfin_kodi/jellyfin/connection_manager.py
@@ -283,21 +283,22 @@ class ConnectionManager(object):
         return None
 
     def _normalize_address(self, address):
-        # TODO: Try HTTPS first, then HTTP if that fails.
-        if '://' not in address:
-            address = 'http://' + address
-
         # Attempt to correct bad input
         url = urllib3.util.parse_url(address.strip())
 
-        if url.scheme is None:
-            url = url._replace(scheme='http')
+        # Default to using https
+        url = url._replace(scheme='https')
 
-        if url.scheme == 'http' and url.port == 80:
-            url = url._replace(port=None)
-
-        if url.scheme == 'https' and url.port == 443:
-            url = url._replace(port=None)
+        # Test if server is reachable over https
+        if self.API.get_public_info(url):
+            if url.scheme == 'https' and url.port == 443:
+                url = url._replace(port=None)
+        else:
+            # Server didn't give an expected response over https
+            # Try http instead
+            url._replace(scheme='http')
+            if url.scheme == 'http' and url.port == 80:
+                url = url._replace(port=None)
 
         return url.url
 
--- a/jellyfin_kodi/jellyfin/connection_manager.py
+++ b/jellyfin_kodi/jellyfin/connection_manager.py
@@ -291,6 +291,7 @@ class ConnectionManager(object):
         url = urllib3.util.parse_url(address.strip())
 
          # Default to using https
+        LOG.info("Attempting HTTPS connection")
         url = url._replace(scheme='https')
       
         # Test if server is reachable over https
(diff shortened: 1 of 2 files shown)


jellyfin_kodi/jellyfin/connection_manager.py
```diff
@@ -283,21 +283,22 @@ def _convert_endpoint_address_to_manual_address(self, info):
...
-        if url.scheme == 'http' and url.port == 80:
-            url = url._replace(port=None)
+        # Default to using https
+        url = url._replace(scheme='https')
 
-        if url.scheme == 'https' and url.port == 443:
-            url = url._replace(port=None)
+        # Test if server is reachable over https
+        if self.API.get_public_info(url):
+            if url.scheme == 'https' and url.port == 443:
```
mcarlton00: Is the port stuff needed anymore?  Since it's just integrated with the URL field, it may not even need a check here

oddstr13: I mostly did it this way to avoid urls such as http://foo:80/asdf, there's no practical *need* for it, just a log readability concern

mcarlton00:
I don't think the fallback to http is working.  Particularly if you have a reverse proxy where some have ssl and some don't.  At least with nginx, it ends up defaulting to using a different certificate since the request came in over 443.

```
2020-03-19 17:21:02.736 T:140175862986496  NOTICE: JELLYFIN.jellyfin.api -> INFO::jellyfin_kodi/jellyfin/api.py:385 Sending get request to system/info/public
2020-03-19 17:21:02.752 T:140175862986496   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <class 'requests.exceptions.SSLError'>
                                            Error Contents: HTTPSConnectionPool(host='jf-test.example.com', port=443): Max retries exceeded with url: /system/info/public (Caused by SSLError(CertificateError("hostname 'jf-test.example.com' doesn't match 'other-domain.example.com'",),))
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/dialogs/servermanual.py", line 84, in onClick
                                                elif self._connect_to_server(server):
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/dialogs/servermanual.py", line 121, in _connect_to_server
                                                result = self.connect_manager.connect_to_address(server)
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/jellyfin/connection_manager.py", line 133, in connect_to_address
                                                address = self._normalize_address(address)
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/jellyfin/connection_manager.py", line 293, in _normalize_address
                                                if self.API.get_public_info(url):
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/jellyfin/api.py", line 432, in get_public_info
                                                response = self.send_request(server_address, "system/info/public")
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/jellyfin/api.py", line 389, in send_request
                                                return request_method(url, **request_settings)
                                              File "/home/matt/.kodi/addons/script.module.requests/lib/requests/api.py", line 75, in get
                                                return request('get', url, params=params, **kwargs)
                                              File "/home/matt/.kodi/addons/script.module.requests/lib/requests/api.py", line 60, in request
                                                return session.request(method=method, url=url, **kwargs)
                                              File "/home/matt/.kodi/addons/script.module.requests/lib/requests/sessions.py", line 533, in request
                                                resp = self.send(prep, **send_kwargs)
                                              File "/home/matt/.kodi/addons/script.module.requests/lib/requests/sessions.py", line 646, in send
                                                r = adapter.send(request, **kwargs)
                                              File "/home/matt/.kodi/addons/script.module.requests/lib/requests/adapters.py", line 514, in send
                                                raise SSLError(e, request=request)
                                            SSLError: HTTPSConnectionPool(host='jf-test.example.com', port=443): Max retries exceeded with url: /system/info/public (Caused by SSLError(CertificateError("hostname 'jf-test.example.com' doesn't match 'other-domain.example.com'",),))
```

mcarlton00:
That fixes the error I saw.

I wonder if we should put an extra log message or two in.  Along the lines of "Attempting secure connection" and "Falling back to insecure http" otherwise users are likely to just see a connection error in the logs and send issues.  It's kinda like the libraries error that gets thrown up on a fresh install.  Thoughts?

mcarlton00:
Actually, I think I have a cleaner way.

https://github.com/jellyfin/jellyfin-kodi/blob/011186edd7c4193428f073b731dce29a39be40af/jellyfin_kodi/jellyfin/api.py#L431-L433

Instead of immediately returning the response here, we could do something involving the `response.history` to determine if a redirect happened.

```python
def get_public_info(self, server_address): 
     response = self.send_request(server_address, "system/info/public") 
    if response.history:
        if response.history[0].response_code == 301
            # make new request with https
    return response.json() if response.status_code == 200 else {}

```

We'll probably have to adjust the response so that the `url` object in connection_manager can be updated correctly as well

TrueTechy:
The addon would still store the address as https with this method though. Using the current implementation it changes the stored url. Also that won't catch the ssl errors we've seen

mcarlton00:
Yes, that's part of the "adjust the response so the `url` object could be updated" that I mentioned.  And it wouldn't need to catch the SSL error that I got earlier because it would never have tried https first.  Following the redirect history allows us to try http, then move to https if the redirect happens.

I don't like the idea of forcing https first and throwing errors.  We've been trying to get rid of the needless errors in the logs, plus it feels too much like using try/except for program flow.

PR #236, opened 2020-03-19
Fix ascii error on context menu

### LINKED ISSUE #220: 0.5.0 init issue on windows / and ascii error
**Describe the bug**
The add-on don't start properly on windows. 
On a fresh kodi install (AppData clean), the add-on first configuration works (server/credentials/library/first sync), then on restart, no "Welcome me!" message and no sync.
Tested on Kodi Windows UWP 18.5 and Kodi 18.6 x64 standard (fresh install in between).
On Android (no fresh install yet), we have the welcome message but no sync appears.

**Logs**
Windows UWP:
https://pastebin.com/7sis95qF
Windows standard x64 Bin:
https://pastebin.com/CBxZHwB0
Android:
https://pastebin.com/PvHdyhhX

**System**
 - OS: [Windows]
 - Jellyfin Version: [10.4.3]
 - Kodi Version: [18.5-6]
 - Addon Version: [0.5.0]

mcarlton00:
Fixes the ascii decoding error when selecting an option with special characters in the context menu of an item.  It seems as though Kodi's internal getLabel return true strings and not respecting the `unicode_literals` in this file

Fixes #220 (again)

### DIFF
--- a/jellyfin_kodi/dialogs/context.py
+++ b/jellyfin_kodi/dialogs/context.py
@@ -64,7 +64,7 @@ class ContextMenu(xbmcgui.WindowXMLDialog):
 
             if self.getFocusId() == LIST:
                 option = self.list_.getSelectedItem()
-                self.selected_option = option.getLabel()
+                self.selected_option = option.getLabel().decode('utf-8')
                 LOG.info('option selected: %s', self.selected_option)
 
                 self.close()
--- a/jellyfin_kodi/dialogs/context.py
+++ b/jellyfin_kodi/dialogs/context.py
@@ -7,6 +7,7 @@ import logging
 import os
 
 from kodi_six import xbmcgui, xbmcaddon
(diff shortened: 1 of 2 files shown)


jellyfin_kodi/dialogs/context.py
```diff
@@ -64,7 +64,7 @@ def onAction(self, action):
 
             if self.getFocusId() == LIST:
                 option = self.list_.getSelectedItem()
-                self.selected_option = option.getLabel()
+                self.selected_option = option.getLabel().decode('utf-8')
```
oddstr13: We should be using `six.ensure_text()` in stead, this will fail if `.getLabel()` ever returns an `unicode` object.
We should also add `version="1.13.0"` to the six dependency, as this function got added somewhere between `v1.11 - 1.13`.

PR #263, opened 2020-04-02
Update deprecated kodi api functions

oddstr13:
Kodi 19 now with less stack traces.


oddstr13:
Yes, I'm running Kodi nightlies on my system, and this fixes a few deprecated (and now removed) Kodi API calls, and a Py2->3 one.

PR #276, opened 2020-04-16
Add badges to README

oddstr13:
I essentially just added all the badges I found that I thought kinda made sense.
I have not paid much attention to where the badges are positioned in relation to each other.

Feedback wanted.

See preview here:
https://github.com/oddstr13/jellyfin-kodi/blob/pr-readme-badges-1/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,17 +10,12 @@
 <a href="https://github.com/jellyfin/jellyfin-kodi">
 <img src="https://img.shields.io/github/license/jellyfin/jellyfin-kodi" alt="GPL 3.0 License" />
 </a>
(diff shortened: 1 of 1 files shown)


thornbill:
The h1 and badges are centered in the other repos. May want to do the same here. https://raw.githubusercontent.com/jellyfin/jellyfin/master/README.md

mcarlton00:
I'm not sure we need most of the GH stuff there.  Issues/PRs/Stars/Last commit are all available on the top of the page.  Forks/code size/commits since release aren't really useful metrics here (imo), but I'm open to having my mind changed.

PR #277, opened 2020-04-19
Attempt to improve logging

oddstr13:


context.py
```diff
@@ -19,10 +18,11 @@
 #################################################################################################
 
 from entrypoint import Context  # noqa: F402
+from helper import LazyLogger  # noqa: F402
 
 #################################################################################################
 
-LOG = logging.getLogger("JELLYFIN.context")
+LOG = LazyLogger("context")
```
mcarlton00, line 25: Should we make all of them `__name__`?  I don't understand giving the root files a special header while everything under jellyfin_kodi uses `__name__`

oddstr13: Good point

PR #289, opened 2020-04-25
Report all items in a playlist

### LINKED ISSUE #198: Playback reporting inconsistently failing
**Describe the bug**

It's currently only possible to control Kodi playback from the web UI when playback is started in the web UI.  If a user starts playback locally via keyboard or remote or such, the server won't properly connect to the running session.

This also has the side effect of losing control when sending music tracks.  The first track will report properly and be controllable, but during subsequent tracks the web UI can't connect to the playing session.

**To Reproduce**

1. Start playback locally
2. From web UI or app, select Kodi from the "Play to" menu
3. No control options
4. Send a new file to Kodi from the control point
5. Control works as expected

Alternative:
1. Send a music album/playlist
2. First song is controllable
3. Following songs are not

 - Kodi Version: 18.3-19alpha
 - Addon Version: likely all?

**Additional context**

Issue seems to stem from https://github.com/jellyfin/jellyfin-kodi/blob/5fc60fce6be8d767d6da93eccca68ba23a61a268/jellyfin_kodi/player.py#L71-#L85.

`window('jellyfin_play.json')` is returning nothing, so the thread never reaches reporting status.

mcarlton00:
Fixes #198 by adding the entire playlist, not just the first track

### DIFF
--- a/jellyfin_kodi/objects/actions.py
+++ b/jellyfin_kodi/objects/actions.py
@@ -215,6 +215,12 @@ class Actions(object):
                 server_address, item['Id'], token)
             listitem.setPath(path)
 
+            play = playutils.PlayUtils(item, False, self.server_id, self.server)
+            source = play.select_source(play.get_sources())
+            play.set_external_subs(source, listitem)
+
+            playutils.set_properties(item, item['PlaybackInfo']['Method'], self.server_id)
+
             playlist.add(path, listitem, index)
             index += 1
 


oddstr13:
Music syncing does not work for Kodi 19, so I am unable to properly test this patch.
I did figure out that I can play music by casting from web however.

Play/Pause & Stop buttons are available for the first item when doing this (same as with watching a single episode / movie), but the controls disappear on subsequent items.

If this patch improves the behaviour on Kodi 18, I'd consider it a *partial* fix for #198.

mcarlton00:
Syncing shouldn't matter here.  Definitely fixes the issue on 18, I tested on both music albums and tv seasons.  That doesn't make sense that Kodi 19 wouldn't be working with this

mcarlton00:
Just tested on a freshly compiled Kodi 19 and it seemed to work fine there as well

PR #290, opened 2020-04-26
Upgrade websocket

### LINKED ISSUE #151: Websockets not working with IPv6
There seems to be a bit of issue with connecting to Jellyfin websockets. Currently for me it's working correctly only for IPv4.

Also addon seems to be connecting to different websocket endpoint compared to web interface.

In web it's ```ws://SERVER_NAME/socket``` and Kodi addon reaches to ```ws://SERVER_NAME/embywebsocket```

**Logs**
```
NOTICE: JELLYFIN.jellyfin.core.ws_client -> INFO::resources/lib/jellyfin/core/ws_client.py:48 Websocket url: ws://{jellyfin-server}/embywebsocket?api_key={jellyfin-token}&device_id=***
NOTICE: JELLYFIN.jellyfin.core.ws_client -> ERROR::resources/lib/jellyfin/core/ws_client.py:65 [Errno -9] Address family for hostname not supported
```

 - OS: Debian 10
 - Jellyfin Version: 10.4.2
 - Kodi Version: 18.4
 - Addon Version: 0.3.6

mcarlton00:
Fully fixes #151 

Upgrades the embedded websocket library to the latest release (0.57), which requires changing the syntax of how we call it.  Rather hacky, but the alternative is repackaging upstream into a compatible Kodi package.  Also requires us patching upstream to include our changes in #274 and #275 (included here in `_http.py`)

Need to decide whether we want to go this route or figure out repacking/patching upstream


oddstr13:
I would prefer to have the websocket library in a separate repository, and pull it in as a dependency.

We should at least look into what would be required to both maintain a fork of the library (if upstream isn't currently accepting patches), and create a script dependency from that repo to push to our Kodi repository.

Ideally, we would get the upstream websocket library into the official Kodi repositories, but I think that is unlikely to happen for now.

oddstr13:
It is also worth noting here for future reference that upstream is [websocket-client/websocket-client](https://github.com/websocket-client/websocket-client) *([PyPi](https://pypi.org/project/websocket_client/))*.

And that the version that is currently embedded is from early 2014 some time (with a few patches applied).

mcarlton00:
Yeah, I'm not super thrilled about the idea of embedding the library in the repo like this, but upstream isn't super active and it saves having to repackage + maintain it in 2 separate repos (kodi 18/19).  I'm not opposed to going that direction, but this is definitely more of a hacky/temporary solution while we figure out the infrastructure bits around that.

mcarlton00:
Closing in favor of #328

PR #298, opened 2020-05-16
Yes.

oddstr13:


mcarlton00:
No :cry: 

2020-05-16-114257_609x847_scrot

oddstr13:
I honestly don't care if it is possible to get the addon installed on older versions of Kodi.
The dependency entry is only to specify minimum compatible version of the API (which they broke by requiring the version number to be `== "3.0.0"`)
If the Kodi team can be bothered spending time to patch this workaround, they *really* should invest time to implement the version dependency system properly.

I'd prefer if this pr at least be kept open for now, to document one possible workaround.
Feel free to add the blocked label.

mcarlton00:
Replaced by #336

PR #301, opened 2020-05-20
Add reset password dialog to addon menu

### LINKED ISSUE #258: Provide method for user to update password
**Describe the bug**

Currently the only way to update a user's password to Jellyfin is to reset the database and do the whole sync again.  We should change add a new dialog in the settings menu to allow this.

mcarlton00:
This dialog only works if Kodi was running when the password was reset in the Web UI.  ~~If Kodi is turned off when the password is reset, then it will automatically let you reset it by starting/stopping/starting Kodi again.  I wanted to catch that so it doesn't need the extra restart cycle, but it's pretty deep and there's not a super good way to verify it's a bad password vs the server just being offline that I can see right now.~~

Fixes #258

Edit: fixed in later commit

### DIFF
--- a/jellyfin_kodi/connect.py
+++ b/jellyfin_kodi/connect.py
@@ -119,6 +119,10 @@ class Connect(object):
                 if 'ExchangeToken' not in state['Servers'][0]:
                     self.login()
 
+            elif state['State'] == CONNECTION_STATE['Unavailable'] and state['Status_Code'] == 401:
+                # If the saved credentials don't work, restart the addon to force the password dialog to open
+                window('jellyfin.restart', clear=True)
+
             elif state['State'] == CONNECTION_STATE['Unavailable']:
                 raise HTTPException('ServerUnreachable', {})
 
--- a/jellyfin_kodi/connect.py
+++ b/jellyfin_kodi/connect.py
@@ -253,8 +253,9 @@ class Connect(object):
         client.set_credentials(get_credentials())
         manager = client.auth
 
+        username = settings('username')
         try:
-            self.login_manual(manager=manager)
+            self.login_manual(user=username, manager=manager)
         except RuntimeError:
             return
 
--- a/jellyfin_kodi/entrypoint/default.py
+++ b/jellyfin_kodi/entrypoint/default.py
@@ -112,6 +112,8 @@ class Events(object):
             xbmc.executebuiltin('Addon.OpenSettings(plugin.video.jellyfin)')
         elif mode == 'adduser':
             add_user()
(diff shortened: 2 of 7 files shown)


TrueTechy:
Could you not pull the public info to see that the servers up? The server response should say about it being an invalid username/password though not sure if that gets passed back to the calling function or if it's just in the logs

mcarlton00:
You asked for this rabbit hole.

The actual response that I care about is happening here: https://github.com/jellyfin/jellyfin-kodi/blob/3f63b3d0e0156b99c1f749669050a974ac37d290/jellyfin_kodi/jellyfin/api.py#L433

That's returned back to here as `system_info`
https://github.com/jellyfin/jellyfin-kodi/blob/3f63b3d0e0156b99c1f749669050a974ac37d290/jellyfin_kodi/jellyfin/connection_manager.py#L310-L322

In this case (failed login) it's just returned as `{ State: 0 }` to https://github.com/jellyfin/jellyfin-kodi/blob/3f63b3d0e0156b99c1f749669050a974ac37d290/jellyfin_kodi/jellyfin/connection_manager.py#L168

Which then gets passed to `connection_manager.py`'s `connect()` function, then `client.py`'s `authenticate()` function, and then finally it gets returned here as `state`
https://github.com/jellyfin/jellyfin-kodi/blob/3f63b3d0e0156b99c1f749669050a974ac37d290/jellyfin_kodi/connect.py#L86-L96

In theory we can capture the 401 status code at the very beginning, but that's a lot of cascading changes that have to happen.  Which is why I just committed what I had at this point.  I'm not sure if that's a path I want to go down at this point

mcarlton00:
Found out we can pass just the status code back up from `api.py` and bypass most of the other functions.  From here we can check if it's a 401 and restart the addon to force the password dialog box to show up.

dkanada:
Is this a password reset or a password change?

mcarlton00:
It's for updating the password in Kodi after it's been changed in the server.  There wasn't any straightforward method to do that before without resetting the whole database and re-syncing the libraries.

TrueTechy:
All looked fine when I tested it. Do we want to show a message to the user so that they know the login failed rather than just throwing the login box at them, as that'll undoubtedly lead to "bug" reports?

mcarlton00:
It will only ever occur when the user changes their password in the server beforehand, so I can't see much benefit in reiterating that they've already done that.

TrueTechy:
> It will only ever occur when the user changes their password in the server beforehand, so I can't see much benefit in reiterating that they've already done that.

On your head be it

PR #308, opened 2020-05-28
Update deprecated isPassword and line1 function parameters

oddstr13:
The isPassword change likely bumps minimum version up to Kodi 18.
This can be worked around if desirable.

### DIFF
--- a/jellyfin_kodi/client.py
+++ b/jellyfin_kodi/client.py
@@ -112,7 +112,7 @@ def reset_device_id():
 
     window('jellyfin_deviceId', clear=True)
     get_device_id(True)
-    dialog("ok", heading="{jellyfin}", line1=translate(33033))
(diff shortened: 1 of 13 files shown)


mcarlton00:
I know we have at least one user on Matrix who was having some instability on Kodi 18 and has rolled back to Kodi 17.  They said that when using multiple profiles that plays were registering on the wrong profile somehow.  We should probably do some testing around this before outright ending Kodi 17 support (even if we haven't purposefully been supporting it up to this point.

The `isPassword` check could likely be handled via a simple version check, but we should at least investigate the profiles issue and preferably not support old versions

mcarlton00:
Issues reported on matrix seem to be related to instability in LibreElec on the user's hardware, nothing to do with us.

Issue #309, opened 2020-05-30
Choose the audio stream and the subtitle stream remotely.


Lumenol:
**Describe the bug**

I cannot choose the audio stream or the subtitles using a remote interface. I can see that it is used but if I change it it is not taken into account by kodi.

**To Reproduce**

1. Activate the remote control of kodi.
2. Start a movie with multiple audio streams or subtitles.
3. Tried to change the audio or subtitle language from the remote interface.
4. Observe that change does not happen.

**Expected behavior**

When I change the audio stream or the subtitles from a remote interface, the change is done in kodi.

**Logs**

```
2020-05-30 18:29:02.112 T:16656  NOTICE: JELLYFIN.player -> INFO::jellyfin_kodi\player.py:161 Setting audio: 3 subs: None
2020-05-30 18:29:02.117 T:16656   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.TypeError'>
                                            Error Contents: unsupported operand type(s) for -: 'unicode' and 'int'
                                            Traceback (most recent call last):
                                              File "C:\Users\crede\AppData\Roaming\Kodi\addons\plugin.video.jellyfin\jellyfin_kodi\monitor.py", line 261, in onNotification
                                                self.general_commands(data)
                                              File "C:\Users\crede\AppData\Roaming\Kodi\addons\plugin.video.jellyfin\jellyfin_kodi\monitor.py", line 388, in general_commands
                                                self.player.set_audio_subs(args['Index'])
                                              File "C:\Users\crede\AppData\Roaming\Kodi\addons\plugin.video.jellyfin\jellyfin_kodi\player.py", line 170, in set_audio_subs
                                                self.setAudioStream(audio - 1)
                                            TypeError: unsupported operand type(s) for -: 'unicode' and 'int'
                                            -->End of Python script error report<--
2020-05-30 18:29:04.413 T:16656  NOTICE: JELLYFIN.player -> INFO::jellyfin_kodi\player.py:161 Setting audio: None subs: 6
2020-05-30 18:29:04.413 T:16656   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
[...]

PR #310, opened 2020-05-30
Fix choose the audio stream and the subtitle stream remotely

Lumenol:
Partially solves the problem #309 
However the poster in the remote interface does not update the first time when I activate or deactivate the subtitles. And the modifications in Koni do not go up in the remote interface.
If I trigger an event on the player like putting it on pause, the information is updated. For what is modifications in Kodi and for activation / deactivation in the remote interface.

I have not done any tests on other versions of Kodi and Jellyfin, so I don't know if my changes are compatible with other versions.

### DIFF
--- a/jellyfin_kodi/monitor.py
+++ b/jellyfin_kodi/monitor.py
@@ -389,6 +389,8 @@ class Monitor(xbmc.Monitor):
             elif command == 'SetSubtitleStreamIndex':
                 self.player.set_audio_subs(None, args['Index'])
 
+            # Kodi needs a bit of time to update it's current status
+            xbmc.sleep(500)
             self.player.report_playback()
 
         elif command == 'DisplayMessage':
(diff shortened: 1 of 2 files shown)


mcarlton00:
Good news/bad news.

Good news: This does indeed fix the initial issue of subs not being changed by the web UI.  Good catch and fix.

Bad news: I'm also confirming the weirdness about the web UI not updating until after it's been clicked again.

Helpful-ish news: I know where it's failing, I just don't know why.  https://github.com/jellyfin/jellyfin-kodi/blob/d50283b84fbdd7cd627beba2ba94dfd87f61ae4a/jellyfin_kodi/player.py#L312

This line is still pulling the old sub index instead of the one that was just updated.  So naturally when it works through the rest of this function and reports status back to the server, it identifies as still using the old sub track.  There's probably another function call needed to trigger that, or possibly it's just a time based thing and we need to wait for the next tick inside of Kodi before it updates itself.  I don't know for sure without digging more

mcarlton00:
I think found a solution.  @Lumenol try this and see if it works for you.

Above this line:
https://github.com/jellyfin/jellyfin-kodi/blob/22677edc052b4e96b6a39c37187d1a334a6ee1cc/jellyfin_kodi/monitor.py#L392

put `xbmc.sleep(500)`. So it will end up looking like:
```python
xbmc.sleep(500)
self.player.report_playback()
```

Basically, Kodi just needs fractions of a second to reach it's next internal tick and update the current item's status.

PR #328, opened 2020-06-21
Remove websocket library

### LINKED ISSUE #151: Websockets not working with IPv6

mcarlton00:
Websocket has been moved to it's own addon we can treat as a dependency in our repo.  This removes the local copy and updates the syntax we're calling it with to use the updated version.

Fixes #151 properly this time


mcarlton00:
Note: On hold until we find a solution to the kodi 18/19 repo bits

oddstr13:
It is also worth noting here for future reference that upstream is [websocket-client/websocket-client](https://github.com/websocket-client/websocket-client) *([PyPi](https://pypi.org/project/websocket_client/))*.

And that the version that is currently embedded is from early 2014 some time (with a few patches applied).

_Originally posted by @oddstr13 in https://github.com/jellyfin/jellyfin-kodi/issues/290#issuecomment-619539006_

MetalGuardian:
any progress on this? without this websocket client update library sync is not working on ipv6 only systems

mcarlton00:
It's in a bit of a grey area, unfortunately, and we haven't fully decided how to handle it all.  I'll try to summarize some of the decisions/issues.

The first issue is that we're kinda duplicating work here, since as part of the MediaImport project somebody else has created a websocket library and the one that I've added to the Jellyfin addon repo isn't going to be needed.  The hangup is that the one in Kodi's repos is only targeting python 3, which won't work on Kodi 18.  So my addon kinda is needed after all.  (From a cursory search I now see that a third person has taken up this, so now there's 3 options for which one to depend on that might get complicated.  Or might solve this particular issue and make it far easier to deal with.  Currently unknown).

Other issues, in no particular order:
* We kinda just hate this library.  It does a whole lot of excess things we don't need, it doesn't follow best practices, and if we're going to be brutally honest they're flat out doing things wrong (which is why I had to change the `ws_client.py`).
* [Upstream](https://github.com/websocket-client/websocket-client) is basically abandonware.  The last commit was in December of 2019, and the maintainer doesn't appear to have been online at all since then.
* There appears to be some debate about the legitimacy of the license of the upstream library, which is more than a little concerning
* The structure of upstream has changed _drastically_, so it's hard to cherry pick the fixes that we need into our local embedded copy

In short:

mcarlton00:
We should be able to move forward with this now.  Upstream has a new maintainer and is moving again and additionally has has began packaging it for Kodi 19 https://github.com/pkscout/script.module.websocket

TrueTechy (approved):

Issue #33, opened 2019-06-23
Unable to play BDMV/DVD files


yangtfu:
all my bdmv files cannot be played by jellyfin plugin, but they can be played via smb path.
This is the same problem i'm facing:
https://emby.media/community/index.php?/topic/68437-bdmv-rips-not-playing-on-emby-402-with-embykodi-beta-add-on/

Does anyone know if it's a bug or configuration issue?

oddstr13:
I don't have any bdmv to test with.
Does it play in the web interface? If not, it's an issue for the server https://github.com/jellyfin/jellyfin/issues
If it does play in web, but not Kodi, that at least narrows down the possibilities some.

I'd suggest remuxing to mkv, but that's probably not practical or desirable for you.

yangtfu:
> I don't have any bdmv to test with.
> Does it play in the web interface? If not, it's an issue for the server https://github.com/jellyfin/jellyfin/issues
> If it does play in web, but not Kodi, that at least narrows down the possibilities some.
> 
> I'd suggest remuxing to mkv, but that's probably not practical or desirable for you.

It works in the web interface. Plus, i'm using kodi native play method in the add-on settings. And it works if i play index.bdmv directly from file path but not in jellyfin scratch items, so i think it's a plugin path parsing issue.

holow29:
Can replicate; here is my log (copied from #327 )

>  2020-06-16 10:47:17.631 T:15525 NOTICE: VideoPlayer::OpenFile: nfs://192.168.1.142/volume1/films/Bluray/Woodstock (1970)
> 2020-06-16 10:47:17.632 T:15596 NOTICE: Creating InputStream
> 2020-06-16 10:47:17.648 T:15596 NOTICE: Creating Demuxer
> 2020-06-16 10:47:17.650 T:15596 ERROR: Read - Error( -14, read call failed with "<0x15>" )
> 2020-06-16 10:47:17.650 T:15596 ERROR: Open - error probing input format, nfs://192.168.1.142/volume1/films/Bluray/Woodstock (1970)
> 2020-06-16 10:47:17.650 T:15596 ERROR: OpenDemuxStream - Error creating demuxer
> 2020-06-16 10:47:17.650 T:15596 NOTICE: CVideoPlayer::OnExit()
[...]

PR #330, opened 2020-06-27
Add support for BDMV movie directories in native playback mode

YK-Samgo:
Add support for BDMV movies in the bluray directry by pointing the file to **./BDMV/index.bdmv** when using **direct path**
related issue #33 

Originally, kodi can't play the movie synced by jellyfin for kodi with **Playback mode: Native (direct paths)** if the bdmv is in a directry like this:
Casino.Royale.2006.COMPLETE.UHD.BLURAY-COASTER/
.
└── Casino.Royale.2006.COMPLETE.UHD.BLURAY-COASTER
    ├── BDMV
    │   ├── BACKUP
    │   ├── BDJO
    │   ├── CLIPINF
    │   ├── JAR
    │   ├── META
    │   ├── PLAYLIST
    │   ├── STREAM
    │   ├── index.bdmv
    │   └── MovieObject.bdmv
    ├── CERTIFICATE
    │   ├── BACKUP
    │   ├── app.discroot.crt
    │   ├── bu.discroot.crt
    │   └── id.bdmv
    ├── Casino.Royale.2006.COMPLETE.UHD.BLURAY-COASTER.nfo
    ├── fanart.jpg
    ├── landscape.jpg
    ├── logo.png
    └── poster.jpg

Jellyfin server tells jellyfin for kodi only the directry of the bdmv, in this case, **Casino.Royale.2006.COMPLETE.UHD.BLURAY-COASTER**, but kodi needs the 'BDMV/index.bdmv' file, which, in this case, should be **Casino.Royale.2006.COMPLETE.UHD.BLURAY-COASTER/BDMV/index.bdmv**. So just give kodi what it wants.

### DIFF
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -268,8 +268,6 @@ def validate_bluray_dir(path):
 
     path = path + '/BDMV/'
 
-    path = path if os.path.supports_unicode_filenames else path
-
     if not xbmcvfs.exists(path):
         return False
 
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -261,6 +261,23 @@ def validate(path):
     return True
 
 
+def validate_bluray_dir(path):
+
+    ''' Verify if path/BDMV/ is accessible.
+    '''
+
+    path = path + '/BDMV/'
+
+    path = path if os.path.supports_unicode_filenames else path
+
+    if not xbmcvfs.exists(path):
+        return False
+
+    window('jellyfin_pathverified.bool', True)
+
+    return True
+
+
 def values(item, keys):
 
     ''' Grab the values in the item for a list of keys {key},{key1}....
--- a/jellyfin_kodi/objects/movies.py
+++ b/jellyfin_kodi/objects/movies.py
@@ -180,7 +180,7 @@ class Movies(KodiDb):
             if validate_bluray_dir(obj['Path'] + obj['Filename']):
                 obj['Path'] = obj['Path'] + obj['Filename'] + '/BDMV/'
                 obj['Filename'] = 'index.bdmv'
-                LOG.debug("path redirect %s",obj['Path'])
+                LOG.debug("Bluray directry %s",obj['Path'])
 
         else:
             obj['Path'] = "plugin://plugin.video.jellyfin/%s/" % obj['LibraryId']
--- a/jellyfin_kodi/objects/movies.py
+++ b/jellyfin_kodi/objects/movies.py
@@ -8,7 +8,7 @@ from kodi_six.utils import py2_encode
 
 import downloader as server
 from database import jellyfin_db, queries as QUEM
-from helper import api, stop, validate, jellyfin_item, library_check, values, Local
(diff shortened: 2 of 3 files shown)


jellyfin_kodi/helper/utils.py
```diff
@@ -261,6 +261,23 @@ def validate(path):
...
 
 
+def validate_bluray_dir(path):
+
+    ''' Verify if path/BDMV/ is accessible.
+    '''
+
+    path = path + '/BDMV/'
+
+    path = path if os.path.supports_unicode_filenames else path
```
mcarlton00, line 271: Doesn't this evaluate to `path` regardless of whether it supports unicode or not?

jellyfin_kodi/objects/movies.py
```diff
@@ -176,6 +176,11 @@ def get_path_filename(self, obj):
                 raise Exception("Failed to validate path. User stopped.")
 
             obj['Path'] = obj['Path'].replace(obj['Filename'], "")
+            '''check bluray directries and point it to ./BDMV/index.bdmv'''
+            if validate_bluray_dir(obj['Path'] + obj['Filename']):
+                obj['Path'] = obj['Path'] + obj['Filename'] + '/BDMV/'
+                obj['Filename'] = 'index.bdmv'
+                LOG.debug("path redirect %s",obj['Path'])
```
mcarlton00, line 183: Maybe change the log entry to something like
```suggestion
                LOG.debug("Bluray directory %s",obj['Path'])
```
Just so it's a bit more descriptive

jellyfin_kodi/helper/utils.py
YK-Samgo: Yes, you are right.
This line is just copied from **def validate(path)** in v0.5.1 and it works , so I didn't put much attention on it. I saw it removed in current version.

holow29:
Not sure if you want to do this separately or not, but you might consider doing the same for DVD folders to point to VIDEO_TS.IFO inside VIDEO_TS folder. (Such as /folder/Casino Royale/VIDEO_TS/VIDEO_TS.IFO)

mcarlton00:
I'd suggest doing so separately just so it's easier to track down if we ever need to find it again.

@holow29 if you're able to test this change reasonably soon it'd be super helpful (nobody on the dev team is currently set up for native mode/has raw bluray rips).  If you can't then it's not a big deal, I'll merge it anyway.  Just having confirmation before merging is always nice :smile:

holow29:
Hmm so I am trying to test (a bit remotely), and I'm having some difficulty because now I am getting something to play, but it is not the correct movie (but always the same incorrect movie). What is happening is similar to before where I click a video tile and the interface seems to reload briefly (maybe then placing the wrong movie under my 'cursor'); the change is that whereas before nothing would play, now something is playing (but the wrong thing, as I said).

mcarlton00:
Perhaps I've missed something, but I was under the impression that this PR was fixing bluray playback for native mode, and that native mode was only accessible from on the same LAN as your server.  So I'm not entirely sure how you could be testing this remotely.

holow29:
I meant that I am not at my setup at the moment, but others are - so I was trying to remotely walk them through some tests. It's a little difficult this way.

PR #336, opened 2020-07-18
Azure pipeline + revamp release process

mcarlton00:
Note: Repo server needs reconfiguring.  Don't merge until after 10.6 releases and we have time to do that.

* Adds an azure pipeline to handle the release process
* Proposing releases from now on are editing `release.yaml` in the project root which will populate `addon.xml`
* Dependencies are handled by the yaml files in `.config` (one for each python version)
* `addon.xml` is generated dynamically on release

The Copy Files and kodirepo commands on the repository server _should_(?) work, but I haven't tested.  I based it off of the commands in the other projects.


.ci/build.yml
```diff
@@ -0,0 +1,38 @@
...
+    - checkout: self
+      clean: true
+
+    - script: python3 -m pip install --user pyyaml
+      displayName: 'Install PyYaml'
+
+    - script: python3 jellyfin-kodi/.config/generate_xml.py ${{ py_version }}
+      displayName: 'Create ${{ py_version }} addon.xml'
+
+    - script: zip -r plugin.video.jellyfin-${{ py_version }}.zip jellyfin-kodi
```
EraYaN, line 19: Use this: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/archive-files?view=azure-devops

.ci/build.yml
```diff
@@ -0,0 +1,38 @@
...
+        sshEndpoint: repository
+        sourceFolder: '${Agent.BuildDirectory}'
+        contents: 'plugin.video.jellyfin*'
+        targetFolder: '/srv/repository/incoming/kodi'
+
+    - task: SSH@0
+      displayName: 'Add to Kodi repo'
+      inputs:
+        sshEndpoint: repository
+        runOptions: 'inline'
```
EraYaN, line 37: You can probably use the commands runOptions here, but if this works it works.
And this now runs for every build including for PRs, that might not be what we want.

mcarlton00: I don't have a repository set up in my account so I can't confirm it works, but I've updated it to more closely match what's in the server using `commands` instead, and added a condition for the tag

.ci/build.yml
```diff
@@ -0,0 +1,43 @@
...
+        archiveFile: 'plugin.video.jellyfin-${{ py_version }}.zip'
+
+    - task: CopyFilesOverSSH@0
+      displayName: 'Upload to repo server'
+      inputs:
+        sshEndpoint: repository
+        sourceFolder: '${Agent.BuildDirectory}'
+        contents: 'plugin.video.jellyfin-${{ py-version }}.zip'
+        targetFolder: '/srv/repository/incoming/kodi'
+      condition: startsWith(variables['Build.SourceBranch'], 'refs/tags')
```
TrueTechy, line 35: So the pipeline runs on every commit and generate the zip files but it'll only upload the stuff to the server on new tags then?

mcarlton00: Correct.  I'm thinking about having it attach builds to PRs for ease of testing in the future, but this is a first step and gets auto releases working again

.config/generate_xml.py
```diff
@@ -0,0 +1,65 @@
...
+with open('jellyfin-kodi/release.yaml', 'r') as f:
+    data = yaml.safe_load(f)
+
+# Populate xml template
+for dep in deps:
+    ET.SubElement(root.find('requires'), 'import', attrib=dep)
+
+# Update version string
+addon_version = data.get('version')
+root.attrib['version'] = f'{addon_version}-{py_version}'
```
oddstr13, line 52: I believe the Kodi version divider is `~` rather than `-`, but `kodirepo` might take care of that?

.config/generate_xml.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Changelog
+changelog = data.get('changelog')
+for section in root.findall('extension'):
+    news = section.findall('news')
+    if news:
+        news[0].text = f'Changelog:\n{changelog}'
```
oddstr13, line 59: I rather like the `v{version} ({date})` header, how about auto-generating that?

.config/generate_xml.py
mcarlton00: Scanning through my installed addons on 19, it looks to be largely `+matrix.1`, a few `x.y.z.1`, and a few `-matrix.1`.  I'm not sure it matters, but we can change it something else if we want to attempt to be consistent with the majority of others

oddstr13: https://kodi.wiki/index.php?title=Addon.xml#How_versioning_works

https://github.com/oddstr13/python-kodirepo/blob/e857de09218a94a4f9f8efced469ecf6798f765b/kodirepo/__main__.py#L110-L122

https://github.com/xbmc/xbmc/blob/1b4b54e731bf35dd72bd03543e0858feb4f0b85a/xbmc/addons/AddonVersion.cpp#L20L24

`a-zA-Z0-9.+_@~` are the allowed characters for version numbers, so if you've seen `*-matrix`, then that probably wasn't working.

I didn't dig deep enough to figure out whether `kodirepo` re-writes the version in the `addon.xml` file, but I don't think it does, and as such we should change the version string produced here.

I think this would make sense, reserving `~` (`-`) for pre-releases if we ever decide to go that route in the furture (and semver specifies `+` as build metadata, which I think fits here?)
```suggestion
root.attrib['version'] = f'{addon_version}+{py_version}'
```

.ci/build.yml
EraYaN: It will do that (attach to PRs) in so far that the pipeline artifacts are downloadable from the "checks page" when you open the build.

obbardc:
<3

PR #339, opened 2020-07-23
Handle empty list of updates from server

o010b:
Fixes "Failed to retrieve latest content updates. No content updates will be applied until Kodi is restarted. (...)" on every start up.

### DIFF
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -396,7 +396,7 @@ class Library(threading.Thread):
             # Get list of updates from server for synced library types and populate work queues
             result = self.server.jellyfin.get_sync_queue(last_sync, ",".join([ x for x in query_filter ]))
             
-            if result == None:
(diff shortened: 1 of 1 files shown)


jellyfin_kodi/library.py
```diff
@@ -393,12 +393,16 @@ def fast_sync(self):
         query_filter = list(set(filters) - set(include))
 
         try:
+            # Get list of updates from server for synced library types and populate work queues
+            result = self.server.jellyfin.get_sync_queue(last_sync, ",".join([ x for x in query_filter ]))
+            
+            if result == None:
```
oddstr13, line 399:
```suggestion
            if result is None:
```

mcarlton00:
It's worth noting that the error you mentioned _usually_ means there was an error with the Kodi Sync Queue plugin in the server, but that shouldn't make an issue on the client side like it does.  Thanks

PR #348, opened 2020-08-02
Various fixes from sonarqube

TrueTechy:


jellyfin_kodi/helper/api.py
```diff
@@ -32,20 +32,14 @@ def get_playcount(self, played, playcount):
 
     def get_naming(self):
 
-        if self.item['Type'] == 'Episode':
+        if self.item['Type'] == 'Episode'and 'SeriesName' in self.item:
```
oddstr13, line 35: Missing space around `and`

service.py
```diff
@@ -25,7 +25,7 @@
 #################################################################################################
 
 LOG = LazyLogger(__name__)
-DELAY = int(settings('startupDelay') if settings('SyncInstallRunDone.bool') else 4 or 0)
+DELAY = int(settings('startupDelay') if settings('SyncInstallRunDone.bool') else 4)
```
oddstr13, line 28: Does `settings('startupDelay')` always return a number? This kinda looks like a fallback for None or empty string.

In any case, I'd move the `int()` to only wrap the setting
```suggestion
DELAY = int(settings('startupDelay')) if settings('SyncInstallRunDone.bool') else 4
```


```suggestion
DELAY = int(settings('startupDelay') or 0) if settings('SyncInstallRunDone.bool') else 4
```

jellyfin_kodi/helper/utils.py
```diff
@@ -214,8 +214,8 @@ def set_screensaver(value):
 
 class JSONRPC(object):
 
-    version = 1
-    jsonrpc = "2.0"
+    id = 1
+    jsonrpc_version = "2.0"
```
oddstr13, line 218: I remember looking at this change myself, I do not remember why I didn't.

service.py
TrueTechy: The startup delay is set in the addon settings and is limited to being a number

jellyfin_kodi/helper/api.py
TrueTechy: Damnit. Only looked through the code changes 3 times :/

jellyfin_kodi/helper/utils.py
TrueTechy: Below is the only place I could find which referenced these variables and everything in my testing still worked so I'd assume it's all okay

anthonylavado:
test

PR #349, opened 2020-08-02
Metadata syncing rework

### LINKED ISSUE #122: Music content is synced to Kodi from deselected libraries
Discussed on this reddit thread: https://www.reddit.com/r/jellyfin/comments/dfdafp/kodi_addon_ignores_or_syncs_wrong_libraries/

**Describe the bug**
When importing music libraries, the way the addon seems to extract from Jellyfin database results in content not in the selected library being pulled over.

**To Reproduce**
Populate multiple music libraries. Compilation albums seem to be the main contributor to the issue.

```
Library A
\
  - Artist A
  \
    - Album 1
    \ - Title

Library B
  - Album 2
  \
    \  - Artist A - Title
       - Artist B - Title
       - Artist C - Title
```

In Jellyfin for Kodi, select only Library A to be imported.

**Expected behavior**
Kodi should only import content actually included in the Libraries selected.

**Actual behavior**
Artists, Albums and Tracks from content not existing in the selected libraries is synced over to Kodi. In the above example, not only does the track from Artist A on Album 2 get synced, but the entirety of Album 2 is pulled over as well.

When the sync is occurring you can follow in the top-right and see the Artist / Album / Title being copied over. It seems that even if Artist A only appears in one track on an album, it pulls over the entire album e.g. "Bee Gees / Now That's What I Call Music X / [every single track]". As a side note, this may also be causing significant performance loss because the same album is pulled multiple times, once for each unique artist that appears in the album.

 - OS: NVIDIA Shield TV 8.0.1
 - Jellyfin Version: 10.4.0
 - Kodi Version: 18.4
 - Addon Version: 0.3.4

### LINKED ISSUE #347: Music sync is unacceptably slow
With the combination of #344 and #331, music syncing now works properly and doesn't spit out a bunch of errors.  But it's too slow.  I'm 10 minutes into a sync and not even out of the A's yet, with a full library of 9000 songs.

Basically, music syncing is likely going to need a full overhaul

mcarlton00:
Started as just polishing music syncing, but ... the codebase forced more changes.  Massively increases music sync speed.

Using a library of 9300 songs:
* 0.5.8: 17.5 minutes
* Post changes: 2.5 minutes

Fixes #347 and fixes #122 

Side effect is that each object type now inherits `library` as an attribute instead of being passed in during each function call.

Example:
```python
obj = Movies(self.server, jellyfindb, videodb, self.direct_path, library)
obj.movie(movie)
```

Also gets rid of the `library_check` wrapper.  This gives _some_ improvement to initial sync times for all media types, but it's not super significant.  Most of the speed improvements come from music syncing only using 3 api calls instead of potentially several thousand depending on library size.


jellyfin_kodi/full_sync.py
```diff
@@ -405,38 +412,42 @@ def music(self, library, dialog):
...
+                    albums = server.get_library_items(library_id, 'MusicAlbum')['Items']
+                    songs = server.get_library_items(library_id, 'Audio')['Items']
+
+                    for index, artist in enumerate(artists):
+                        artist_name = artist.get('Name')
+
+                        # Update percentage dialog
+                        percent = int((float(index) / float(num_artists)) * 100)
+                        message = artist_name
+                        dialog.update(percent, heading="%s: %s" % (translate('addon_name'), library['Name']), message=message)
```
TrueTechy, line 432: Why we directly assigning one variable name to another. Couldn't you just put the `artist_name` in the message value?

jellyfin_kodi/helper/utils.py
```diff
@@ -535,3 +536,18 @@ def get_filesystem_encoding():
         enc = 'utf-8'
 
     return enc
+
+
+def find_library(server, item):
+    from database import get_sync
```
TrueTechy, line 542: Need to move import to top of file instead of in function

jellyfin_kodi/full_sync.py
mcarlton00: Mostly blind copy/paste from the previous code for this part

jellyfin_kodi/helper/utils.py
mcarlton00: It doesn't work.  No, I don't know why, but it throws import errors in the database init.  Mystery for another day, imo.

TrueTechy:
I got the "error with sync" dialog come up doing a sync of all the libraries. 

```
2020-08-03 19:59:41.585 T:139694725646080  NOTICE: JELLYFIN.full_sync -> ERROR::jellyfin_kodi/full_sync.py:256 full sync exited unexpectedly
2020-08-03 19:59:41.588 T:139694725646080  NOTICE: JELLYFIN.full_sync -> ERROR::jellyfin_kodi/full_sync.py:257 local variable 'boxset_library' referenced before assignment
                                              File "jellyfin_kodi/full_sync.py", line 223, in process_library
                                                if boxset_library:
                                            UnboundLocalError: local variable 'boxset_library' referenced before assignment
2020-08-03 19:59:41.595 T:139694725646080  NOTICE: JELLYFIN.full_sync -> INFO::jellyfin_kodi/full_sync.py:579 --<[ fullsync ]
2020-08-03 19:59:41.598 T:139694725646080  NOTICE: JELLYFIN.library -> ERROR::jellyfin_kodi/library.py:367 local variable 'boxset_library' referenced before assignment
                                              File "jellyfin_kodi/library.py", line 324, in startup
                                                sync.libraries()
                                              File "jellyfin_kodi/full_sync.py", line 101, in libraries
                                                self.start()
                                              File "jellyfin_kodi/full_sync.py", line 186, in start
                                                self.process_library(library)
                                              File "jellyfin_kodi/full_sync.py", line 223, in process_library
                                                if boxset_library:
                                            UnboundLocalError: local variable 'boxset_library' referenced before assignment
```

Looks like that section of code might just need indented one more level

```
if library_id.startswith('Boxsets:'):

                libraries = self.get_libraries(library_id.split('Boxsets:')[1] if len(library_id) > len('Boxsets:') else None)
                for entry in libraries:
                    if entry[2] == 'boxsets':
                        boxset_library = {'Id': entry[0], 'Name': entry[1]}
                        break

                if boxset_library:
                    if library_id.endswith('Refresh'):
                        self.refresh_boxsets(boxset_library)
                    else:
                        self.boxsets(boxset_library)

                return
```

mcarlton00:
It was throwing errors if a server doesn't have any collections defined.  good catch

PR #362, opened 2020-08-13
Refactor metadata ancestor gathering

### LINKED ISSUE #361: Incremental sync very slow
While #349 made pretty significant improvements to initial sync speeds, it's slowed down the incremental sync.  By calling `find_library()` on every update, it's taking 1-2 seconds for that function call for each updated item since kodi was last started.  Seems to be a bit worse with music items than video items.

mcarlton00:
Kinda fixes #361. Some of it just can't be helped.  We don't need to find ancestors on userdata updates.  If the item exists, it's updated. If it doesn't, it gets skipped.  No need to do the intensive lookup process.

I'm not sure the Artist/Album bits of the userdata updates are ever triggered.  At least I haven't seen it in my testing, but I'm not confident enough to remove it yet.

Also changes
```
Music(*default_args).song(item)
```
to
```
Music(*default_args).userdata(item)
```
We don't want to try to add new items whenever a song is listened to

### DIFF
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -672,28 +672,25 @@ class UserDataWorker(threading.Thread):
                 except Queue.Empty:
                     break
 
-                # Verify that the updated item is in our local whitelist
-                library = find_library(self.server, item)
-                if library:
-                    default_args = (self.server, jellyfindb, kodidb, self.direct_path, library)
-                    try:
-                        if item['Type'] == 'Movie':
-                            Movies(*default_args).userdata(item)
-                        elif item['Type'] in ['Series', 'Season', 'Episode']:
-                            TVShows(*default_args).userdata(item)
-                        elif item['Type'] == 'MusicAlbum':
-                            Music(*default_args).album(item)
-                        elif item['Type'] == 'MusicArtist':
-                            Music(*default_args).artist(item)
-                        elif item['Type'] == 'AlbumArtist':
-                            Music(*default_args).albumartist(item)
-                        elif item['Type'] == 'Audio':
-                            Music(*default_args).song(item)
-                    except LibraryException as error:
-                        if error.status == 'StopCalled':
-                            break
-                    except Exception as error:
-                        LOG.exception(error)
+                default_args = (self.server, jellyfindb, kodidb, self.direct_path)
+                try:
+                    if item['Type'] == 'Movie':
+                        Movies(*default_args).userdata(item)
+                    elif item['Type'] in ['Series', 'Season', 'Episode']:
+                        TVShows(*default_args).userdata(item)
+                    elif item['Type'] == 'MusicAlbum':
+                        Music(*default_args).album(item)
+                    elif item['Type'] == 'MusicArtist':
+                        Music(*default_args).artist(item)
+                    elif item['Type'] == 'AlbumArtist':
+                        Music(*default_args).albumartist(item)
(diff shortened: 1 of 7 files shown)


TrueTechy:
Getting errors trying to do the initial sync 

```
2020-08-14 23:58:38.989 T:140495737046784  NOTICE: JELLYFIN.library -> ERROR::jellyfin_kodi/library.py:367 unsupported operand type(s) for /: 'unicode' and 'float'
                                              File "jellyfin_kodi/library.py", line 324, in startup
                                                sync.libraries()
                                              File "jellyfin_kodi/full_sync.py", line 101, in libraries
                                                self.start()
                                              File "jellyfin_kodi/full_sync.py", line 186, in start
                                                self.process_library(library)
                                              File "jellyfin_kodi/full_sync.py", line 243, in process_library
                                              File "jellyfin_kodi/helper/wrapper.py", line 41, in wrapper
                                              File "jellyfin_kodi/full_sync.py", line 386, in musicvideos
                                                obj.musicvideo(mvideo)
                                              File "jellyfin_kodi/helper/wrapper.py", line 65, in wrapper
                                                return func(*args, **kwargs)
                                              File "jellyfin_kodi/helper/wrapper.py", line 77, in wrapper
                                              File "jellyfin_kodi/objects/musicvideos.py", line 82, in musicvideo
                                                obj['Resume'] = API.adjust_resume((obj['Resume'] or 0) / 10000000.0)
                                            TypeError: unsupported operand type(s) for /: 'unicode' and 'float'
```

mcarlton00:
I'm not sure the referenced error is caused by this PR.  Or at least I can't imagine how it would be.  Stranger things have certainly happened in this code base, though.

Fixes #361 more gooder now.  

Part 1 was moving the `find_library()` function from `library.py` into each individual function.  While it's easier to understand in `library.py`, it's effectively causing 1-2 seconds per item in the incremental sync.  Which is an pretty major annoyance if you run into a case like I had with ~600 items changed.

The `jellyfin_item()` wrapper on each media object looks up the item in the local database and pulls various information from it, including the LibraryId of the object (stores this metadata in the `e_item` variable).  For reasons unknown, LibraryId wasn't ever included for Albums and Audio files.  When the LibraryId (or the DB entry entirely) can't be found, it falls back into the `find_library()` function that checks the server for ancestors and takes 1-2 seconds to complete, so obviously it's better to pull this info from the database instead.

This changes the query so that any _future_ modified Albums/Audio files get the LibraryId added to it's entry in the local database.  However doing it this way means that incremental syncs that touch the un-modified entries are still rather slow.  We should recommend running a manual sync on music libraries so users don't see a big delay more than necessary.

jellyfin_kodi/library.py
```diff
@@ -605,39 +605,37 @@ def run(self):
...
-                        if self.notify:
-                            self.notify_output.put((item['Type'], api.API(item).get_naming()))
-                    except LibraryException as error:
-                        if error.status == 'StopCalled':
-                            break
-                    except Exception as error:
-                        LOG.exception(error)
+                default_args = (self.server, jellyfindb, kodidb, self.direct_path)
+                try:
+                    LOG.info('{} - {}'.format(item['Type'], item['Name']))
```
oddstr13, line 610: Should this really be info? Looks like it would be rather noisy.
```suggestion
                    LOG.debug('{} - {}'.format(item['Type'], item['Name']))
```

oddstr13 (approved):

PR #364, opened 2020-08-15
Removed webservice

TrueTechy:
Cannot determine what the point of this webservice was for. It says about being there for playback but all my testing post removal has shown no issues playing things without it.

It's also not uncommon for this to not work on restricted devices so I'd imagine if there was some functionality breaking use for this we'd of heard of it by now


cvium:
It's for .strm playback, which is a feature that wasn't entirely complete when we forked.

TrueTechy:
Didn't see anything in the code about it pulling in subtitle data. Seemed to just be a different way of kodi pulling the data instead of using the plugin path stuff. Can we not use the standard subtitle mechanisms for strm?

cvium:
strm is not related to subtitles. It's basically a text file with an url (or multiple?) that should be fed to the player.

Issue #380, opened 2020-09-02
Unable to sync music library


nellistc:
**Describe the bug**
Attempt to sync music library fails. Video libraries sync successfully. Music library is able to be browsed as a dynamic library through the addon.

**To Reproduce**
1. Install latest jellyfin addon
2. Connect to jellyfin server
3. Attempt to sync music library
4. Sync fails with 'something went wrong' error. Kodi displays **Gathering: Music** status message and progress stays at 0%.

**Expected behavior**
Music library to be synced

**System (please complete the following information):**
 - OS: Linux
 - Jellyfin Version: 10.6.4
 - Kodi Version: 18.8 (libreelec 9.2.4)
 - Addon Version: 0.5.9-py2
 - Playback Mode: Add-on

**Additional information**

Initially, the request times out. If the DEFAULT_HTTP_TIMEOUT value is changed to a very long time (eg, 3000 seconds), the last line shown in the debug log is

2020-09-02 22:22:28.017 T:139790238344960  NOTICE: JELLYFIN.jellyfin.http -> DEBUG::jellyfin_kodi/jellyfin/http.py:162 ---<[ http ][77127 ms]

mcarlton00:
77 seconds is a _long_ time for that api call.  Mine was only taking roughly 12 seconds to gather everything.  How large is your music library?

nellistc:
It’s approximately 900 albums containing 12,700 songs.

The server is not fast, so this may be a contributing factor. However, it has been synced previously with an older version of kodi and the add on. (I cant be sure which versions.)

nellistc:
The process (at least according to the logs) woke up about 11 hours after that last debug message and now I have the library in sync. Very strange.

mcarlton00:
that's very odd.  My library is roughly 9000 songs and it was fine.  Being libreelec, I'm assuming this is a Pi or similar SBC?  Also was debug logging on when you were doing the sync initially, or did you turn it on after the fact and try again?

0.5.9 saw a pretty massive rework to the music syncing process, which if you've synced on previous versions then you probably understand why it was needed.  Basically, instead of making an api call for every single artist and album like in previous versions, now we just query the server for all artists, albums, and tracks in 3 calls and process them locally.  This causes the longer "gathering" delay waiting for the api to respond with all of those objects, but that time is more than made up in not having to make repeated calls to the api (Saves 15 minutes on my music sync).  #349 had the brunt of that work.

nellistc:
[...]

PR #387, opened 2020-09-08
Optimize music sync for large libraries

### LINKED ISSUE #386: Music Sync fails on extra large libraries
Hi, I'm experiencing a similar issue:
I tried to sync the music-library and after 12:20 min. I received an error-message (Something went wrong).
In this library are about 45'000 songs.
If any log-files would help you for this topic, I would be glad to share them.

_Originally posted by @n-buck in https://github.com/jellyfin/jellyfin-kodi/issues/380#issuecomment-688072472_

Logs: https://paste.kodi.tv/ewodijidiv

mcarlton00:
Instead of pulling all resources from the server in a single api call, utilize the "Paging - Max Items" setting and pull them in batches.  Default value is 15 items per batch.  Loads them with a generator so we can start processing them locally while still loading results from the server.

Changes the sync order a bit because of the way the loading happens.  Instead of processing _all_ of artist A, then moving on to all of artist B, it now syncs all artists, followed by all albums, with all songs at the end.

Slightly slower sync speeds than the previous iteration, but should be much safer when large libraries are concerned, and still much faster than 0.5.8.  Fixes #386 and should prevent another situation similar to #380 from happening.

### DIFF
--- a/jellyfin_kodi/downloader.py
+++ b/jellyfin_kodi/downloader.py
@@ -212,8 +212,6 @@ def get_items(parent_id, item_type=None, basic=False, params=None):
 
 def get_artists(parent_id=None):
 
-    #url = "Artists"
-
     query = {
         'url': 'Artists',
         'params': {
--- a/jellyfin_kodi/downloader.py
+++ b/jellyfin_kodi/downloader.py
@@ -164,6 +164,27 @@ def get_episode_by_season(show_id, season_id):
         yield items
 
 
+def get_item_count(parent_id, item_type=None, params=None):
+
+    url = "Users/{UserId}/Items"
+
+    query_params = {
+        'ParentId': parent_id,
+        'IncludeItemTypes': item_type,
+        'EnableTotalRecordCount': True,
+        'LocationTypes': "FileSystem,Remote,Offline",
+        'Recursive': True,
+        'Limit': 1
+    }
+    if params:
+        query_params['params'].update(params)
+
+    result = _get(url, query_params)
+
+    total = result.get('TotalRecordCount')
+
+    return total
+
 def get_items(parent_id, item_type=None, basic=False, params=None):
 
     query = {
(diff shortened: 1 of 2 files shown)


mcarlton00:
@n-buck if you can, I'd appreciate a test with your giant library to see if this works properly.  Be warned that it's going to take a bit of time.  As a comparison, my 9300 song library takes just over 3 minutes, and it's largely dependent on local device speed.

* [Kodi 18 Version](https://artprodsu6weu.artifacts.visualstudio.com/Aa752b3a7-9db7-4862-a13c-768672f19bbe/7cce6c46-d610-45e3-9fb7-65a6bfd1b671/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2plbGx5ZmluLXByb2plY3QvcHJvamVjdElkLzdjY2U2YzQ2LWQ2MTAtNDVlMy05ZmI3LTY1YTZiZmQxYjY3MS9idWlsZElkLzE0MjMxL2FydGlmYWN0TmFtZS9wbHVnaW4udmlkZW8uamVsbHlmaW4tcHkyLTIwMjAwOTA4LjI1/content?format=zip)
* [Kodi 19 Version](https://artprodsu6weu.artifacts.visualstudio.com/Aa752b3a7-9db7-4862-a13c-768672f19bbe/7cce6c46-d610-45e3-9fb7-65a6bfd1b671/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2plbGx5ZmluLXByb2plY3QvcHJvamVjdElkLzdjY2U2YzQ2LWQ2MTAtNDVlMy05ZmI3LTY1YTZiZmQxYjY3MS9idWlsZElkLzE0MjMxL2FydGlmYWN0TmFtZS9wbHVnaW4udmlkZW8uamVsbHlmaW4tcHkzLTIwMjAwOTA4LjI1/content?format=zip)

niawag:
I've tested this pull for Kodi 18 version (on coreelec, latest stable version) and it's working great. Thanks a lot!

jellyfin_kodi/downloader.py
```diff
@@ -164,6 +164,27 @@ def get_episode_by_season(show_id, season_id):
...
+        'Limit': 1
+    }
+    if params:
+        query_params['params'].update(params)
+
+    result = _get(url, query_params)
+
+    total = result.get('TotalRecordCount')
+
+    return total
```
TrueTechy, line 186: Do we really want to assign things to just return them immediately

jellyfin_kodi/full_sync.py
```diff
@@ -417,37 +417,47 @@ def music(self, library, dialog):
...
+                    artists = server.get_artists(library_id)
+                    for batch in artists:
+                        for item in batch['Items']:
+                            LOG.debug('Artist: {}'.format(item.get('Name')))
+                            percent = int((float(count) / float(total_items)) * 100)
+                            dialog.update(percent, message='Artist: {}'.format(item.get('Name')))
+                            obj.artist(item)
+                            count += 1
+                            # Delete item once it's been processed for memory management
+                            del item
```
TrueTechy, line 439: Does this actually delete anything from the dict to save memory? doesn't it just get replaced with the next item when the loop jumps back to the top?

TrueTechy (approved):

jellyfin_kodi/downloader.py
mcarlton00: I don't have particularly strong feelings about it, but yeah it's not really needed here

jellyfin_kodi/full_sync.py
mcarlton00: My impression was that if we went back and looped through `artists` a second time, all of these items would still be there and accessible again, in which case deleting them should help free memory up.  I haven't explicitly tested that though.

mcarlton00: Test shows that it did not in fact do anything to del the item there.

oddstr13 (approved):
Unless there's a good reason to keep the batch object (anything other than `Items` being used?), I'd prefer if the `for item in batch['Items']` loop was moved a couple of functions up (into `_get_items`).

That's something for another pullrequest however.

n-buck:
@mcarlton00 I just tested it and it works flawlessly with my library.
Thank you for the great and fast work!

PR #395, opened 2020-09-15
Add more transcoding options

Electry:
I made these changes mostly for my personal use, but I'm parking them here in case someone deems them useful, and perhaps could review them and/or help me test them more extensively.

Anyway, this PR adds more transcoding options to the Settings menu, such as preferred transcoding codec for video/audio, manual bitrate options, max audio channels, and few other changes (described in commit messages).

Max. audio bitrate while transcoding/direct streaming seems to be limited by the backend to [384 kbps](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L1370), just like max. channels are [limited to 6](https://github.com/jellyfin/jellyfin/blob/master/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L1459). Transcoding to AC3 could benefit from allowing up to 640 kbps, but that needs to be dealt with on the server side first.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -289,20 +289,58 @@ class PlayUtils(object):
 
         return self.info['Path']
 
-    def get_bitrate(self):
+    def get_max_bitrate(self):
 
         ''' Get the video quality based on add-on settings.
             Max bit rate supported by server: 2147483 (max signed 32bit integer)
         '''
-        bitrate = [664, 996, 1320, 2000, 3200,
-                   4700, 6200, 7700, 9200, 10700,
-                   12200, 13700, 15200, 16700, 18200,
-                   20000, 25000, 30000, 35000, 40000,
-                   100000, 1000000, 2147483]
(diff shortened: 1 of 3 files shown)


mcarlton00:
Thanks, nice work.  More config options are always good, and transcoding definitely doesn't get a ton of love since we generally direct play everything.  A quick glance through the code and it looks fine, but I'll do some testing and more in depth review later this week.

In the meantime, if anybody wants to help test this you can download the built zip files from [here](https://dev.azure.com/jellyfin-project/jellyfin/_build/results?buildId=14699&view=artifacts&type=publishedArtifacts)

mcarlton00:
Just fyi, you're ready for this to be reviewed, mark it as ready instead of a draft

Electry:
Added option to transcode VC-1 (no hw decode support on RPi3 without the license key).

I will make a separate PR for the "Maximum artwork resolution" option, since it isn't really related to transcoding.

mcarlton00 (approved):
Everything seems fine to me.  I haven't seen any errors while testing.  Nice work

PR #399, opened 2020-09-17
Fix extracted subtitle delivery for transcodes

### LINKED ISSUE #189: Kodi Doesnt Display Subtitles
Pinning this from the forums for @tazmanov 

https://forum.jellyfin.org/t/kodi-not-finding-embedded-subtitles/1345/3

Kodi is not detecting any embedded subrip (srt) subtitles.

bsdf:
On-the-fly extracted subtitles have `IsExternal: false` so they will be ignored by the plugin and not made available to Kodi. This PR changes the plugin to check `DeliveryMethod` instead like the WebUI

Possibly fixes #189

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -468,14 +468,9 @@ class PlayUtils(object):
         kodi = 0
 
         for stream in source['MediaStreams']:
-
-            if stream['Type'] == 'Subtitle' and stream['IsExternal']:
+            if stream['Type'] == 'Subtitle' and stream['DeliveryMethod'] == 'External':
                 index = stream['Index']
-
-                if 'DeliveryUrl' in stream and stream['DeliveryUrl'].lower().startswith('/videos'):
-                    url = "%s/%s" % (self.info['ServerAddress'], stream['DeliveryUrl'])
-                else:
-                    url = self.get_subtitles(source, stream, index)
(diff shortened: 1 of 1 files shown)


jellyfin_kodi/helper/playutils.py
```diff
@@ -468,14 +468,9 @@ def set_external_subs(self, source, listitem):
         kodi = 0
 
         for stream in source['MediaStreams']:
-
-            if stream['Type'] == 'Subtitle' and stream['IsExternal']:
+            if stream['Type'] == 'Subtitle' and stream['DeliveryMethod'] == 'External':
                 index = stream['Index']
-
-                if 'DeliveryUrl' in stream and stream['DeliveryUrl'].lower().startswith('/videos'):
-                    url = "%s/%s" % (self.info['ServerAddress'], stream['DeliveryUrl'])
```
bsdf, line 476: this code is already present in `get_subtitles`

mcarlton00 (approved):
Seems to work fine from what I can tell, and less code is always nice

Electry:
With this PR, kodi completely ignores 'Allow subtitle extraction on the fly' server option, causing playback for content with embedded subtitles not to start until they are all extracted (takes ~20 minutes for 1080p remux stored in remote/cloud drive).

Video player OSD (while trasncoding) also contains subtitles that aren't available for extraction, such as embedded PGS. Selecting those doesn't do anything.

PR #402, opened 2020-09-27
Simplify code 🔨

bpepple:
Bunch of fairly minor changes to simplify the code and improve readability.

### DIFF
--- a/jellyfin_kodi/downloader.py
+++ b/jellyfin_kodi/downloader.py
@@ -79,7 +79,7 @@ def validate_view(library_id, item_id):
         LOG.exception(error)
         return False
 
-    return True if len(result['Items']) else False
+    return bool(len(result['Items']))
 
 
 def get_single_item(parent_id, media):
--- a/jellyfin_kodi/downloader.py
+++ b/jellyfin_kodi/downloader.py
@@ -129,8 +129,7 @@ def get_filtered_section(parent_id=None, media=None, limit=None, recursive=None,
 
 def get_movies_by_boxset(boxset_id):
 
-    for items in get_items(boxset_id, "Movie"):
-        yield items
+    yield from get_items(boxset_id, "Movie")
 
 
 def get_episode_by_show(show_id):
@@ -144,8 +143,7 @@ def get_episode_by_show(show_id):
             'Fields': api.info()
         }
     }
-    for items in _get_items(query):
-        yield items
+    yield from _get_items(query)
 
 
 def get_episode_by_season(show_id, season_id):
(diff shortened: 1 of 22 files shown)


jellyfin_kodi/downloader.py
```diff
@@ -129,8 +129,7 @@ def get_filtered_section(parent_id=None, media=None, limit=None, recursive=None,
 
 def get_movies_by_boxset(boxset_id):
 
-    for items in get_items(boxset_id, "Movie"):
-        yield items
```
TrueTechy, line 133: We currently support kodi 17 and 18 which both use Python 2.7, this syntax was introduced in python 3 and so we can't use it

bpepple: Ahh, haven't written any python-2.* code in ages. I believe the yield commits might be the only ones that are python3 specific since the other commits are for boolean & if statement/expression changes.

bpepple: Reverted this commit.

jellyfin_kodi/objects/obj.py
```diff
@@ -125,11 +125,10 @@ def __recursiveloop__(self, obj, keys):
...
 
         if obj:
-            if rest:
-                for item in obj:
+            for item in obj:
+                if rest:
                     self.__recursiveloop__(item, rest)
-            else:
-                for item in obj:
+                else:
```
oddstr13, line 131: This looks like a potential logic change, needs double check (we really need to get good test coverage going.)

oddstr13 (changes requested):
Additionally, this PR would preferably be squashed – many commits with similar or same changes.
Could squash merge if need be.

bpepple:
@oddstr13 Based on your comments, it looks like you're not terribly interested in these changes so it's probably best we reject this pull. And I'll just keep a local fork for the syncing issue I'm looking to fix. Thx!

oddstr13:
About half of it was more than fine, and some of the notes where for a more detailed review of the (potential) functionality changes, just making absolutely sure behavior doesn't change for what's essentially a styling PR.

I don't mind doing the requested changes, verifying flake8 says it's fine and squashing if you wanna just leave the PR open for now.
I might not get around to it right away however.

Keeping flake8 happy is a hard requirement from my side, whether that is done by ignoring specific warnings in a few places (unused imports in `__init__`), changing flake8 settings (should probably be done in separate PRs with discussion about it), or just changing code.

Additionally I'd prefer the [sonarcube numbers](https://sonarcloud.io/dashboard?id=jellyfin_jellyfin-kodi) going down, and many of the lines you refractored could quite easily make them go down. [Example of how code complexity is calculated](https://sonarcloud.io/project/issues?id=jellyfin_jellyfin-kodi&open=AXGAC5hhXZZoTpqcHs7G&resolved=false&severities=CRITICAL&types=CODE_SMELL).

oddstr13:
I've done the git magic now;
* squashed commits together where it made sense
* dropped the websocket changes
* rebased & resolved conflicts

The original PR is available in this branch; https://github.com/oddstr13/jellyfin-kodi/commits/pr-402-backup-original

PR #404, opened 2020-09-29
Enable flake8 and pytest on pullrequest (GitHub Actions)

oddstr13:
Lints fixed separately in #403 

To be replaced by azure pipelines whenever we get around to figuring that out.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,7 +14,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 end_of_line = lf
-max_line_length = null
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -14,7 +14,7 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 end_of_line = lf
-max_line_length = null
+max_line_length = 9999
```
oddstr13, line 17: I think it was pytest? that wanted this to either `== off` or be a int – but off didn't work.

PR #413, opened 2020-10-19
Fix need for restart/reboot for sync settings changes to take effect

macearl:
If `LIMIT` and `DTHREADS` are defined as global variables they only get assigned once at startup, this means that any changes to the number of download threads and max. items only take effect after the plugin gets restarted or the complete system reboots.

By moving them into the function itself they are read every time new items are fetched.

I'm assuming uppercase variable names are reserved for global variables, so i converted them to lowercase as well.


mcarlton00 (approved):
Seems like a good change to me

PR #416, opened 2020-10-29
Byebye TheVoid

cvium:
UNTESTED because I am not sure how to build this anymore...

PR #417, opened 2020-10-29
Don't force-extract subs for transcoded content

Electry:
#399 broke playback of transcoded content with embedded subs. Kodi now waits for all embedded subs to be extracted while ignoring 'Allow subtitle extraction on the fly' server option.

On my instance, the playback won't start for about 20 minutes, until the subs are finished extracting (even if the option is disabled in the server dashboard).

See my comment here: https://github.com/jellyfin/jellyfin-kodi/pull/399#issuecomment-718810945

This PR also adds embedded subs to the list of available sub tracks in the audio/subs selection dialog (the one shown before playback starts), if the aforementioned option is enabled on your jellyfin server. Furthemore, now subs that don't have 'SupportsExternalStream' set to true are also ignored and not shown as external in the video player OSD sub selection (those can't be extracted, if I'm not mistaken) - such as embedded PGS.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -509,8 +509,13 @@ class PlayUtils(object):
         mapping = {}
         kodi = 0
 
+        server_settings = TheVoid('GetTranscodeOptions', {'ServerId': self.info['ServerId']}).get()
+
         for stream in source['MediaStreams']:
-            if stream['Type'] == 'Subtitle' and stream['DeliveryMethod'] == 'External':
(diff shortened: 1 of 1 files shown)


mcarlton00:
Haven't been able to test yet, but seems like a reasonable fix, and it's not something I would have caught, so thanks.

If it's not a big deal to you, I'm going to hold off on merging until after we finish sorting out #416, which will require a couple small changes here.  We just really want ThetVoid gone.

mcarlton00 (approved):
Well, I lied.  TheVoid is going to need more work, so we'll do this one first.  Everything is functioning properly best I can tell, so it looks good.

Issue #428, opened 2020-11-20
Music database changes in Kodi 19


oddstr13:
```python
2020-11-20 11:12:04.823 T:29417    INFO <general>: JELLYFIN.database -> ERROR::jellyfin_kodi/database/__init__.py:163 type: <class 'sqlite3.OperationalError'> value: no such column: strFanart
2020-11-20 11:12:04.839 T:29417    INFO <general>: JELLYFIN.database -> ERROR::jellyfin_kodi/database/__init__.py:163 type: <class 'sqlite3.OperationalError'> value: no such column: strFanart
2020-11-20 11:12:04.852 T:29246    INFO <general>: Loading skin file: DialogConfirm.xml, load type: KEEP_IN_MEMORY
2020-11-20 11:12:13.154 T:29417    INFO <general>: JELLYFIN.full_sync -> ERROR::jellyfin_kodi/full_sync.py:257 full sync exited unexpectedly
2020-11-20 11:12:13.190 T:29417    INFO <general>: JELLYFIN.full_sync -> ERROR::jellyfin_kodi/full_sync.py:258 no such column: strFanart
                                                   Traceback (most recent call last):
                                                     File "jellyfin_kodi/full_sync.py", line 243, in process_library
                                                       media[library['CollectionType']](library)
                                                     File "jellyfin_kodi/helper/wrapper.py", line 41, in wrapper
                                                       result = func(self, dialog=dialog, *args, **kwargs)
                                                     File "jellyfin_kodi/full_sync.py", line 436, in music
                                                       obj.artist(item)
                                                     File "jellyfin_kodi/helper/wrapper.py", line 65, in wrapper
                                                       return func(*args, **kwargs)
                                                     File "jellyfin_kodi/helper/wrapper.py", line 77, in wrapper
                                                       return func(self, item, e_item=e_item, *args, **kwargs)
                                                     File "jellyfin_kodi/objects/music.py", line 93, in artist
                                                       self.update(obj['Genre'], obj['Bio'], obj['Thumb'], obj['Backdrops'], obj['LastScraped'], obj['ArtistId'])
                                                     File "jellyfin_kodi/objects/kodi/music.py", line 91, in update
                                                       self.cursor.execute(QU.update_artist, args)
                                                   sqlite3.OperationalError: no such column: strFanart
                                                   
[...]

PR #462, opened 2021-01-23
Fixes for Matrix-RC2

mcarlton00:
* Fix 'User' label being shown in manual login dialogs (label seemed to be hidden in previous kodi versions and is shown now) ((https://forum.jellyfin.org/t/kodi-addon-broken/3903))
* ~~Fix boxsets not syncing properly (not particularly happy with how this looks, but don't have a better way in mind at this moment)~~ Fixed previously, my branch was outdated
* Fix casting from other clients (likely more Void fallout, probably not matrix related, but fixing it since I'm here)
* Fix music syncing #428


TrueTechy (approved):
LGTM. Tested sync and playback on Kodi 18.9 and 19.something

oddstr13 (approved):

Issue #469, opened 2021-02-12
Sync issue - sqlite integrity error due to duplicate actor name


Josko:
**Describe the bug**
When syncing (or continuing, after it didn't work) between Kodi and Jellyfin, it just hangs indefinitely in Kodi, after reaching some % (e.g. 18%, sometimes 30%). One has to kill kodi-x11 for it to quit. Reinstalling/restarting/native vs addon, nothing really helps.

**To Reproduce**
1. Attempt to sync Jellyfin and Kodi using any or all libraries after it asks (or continue "previous sync")
2. Sync will work partially, anywhere from 0% to 30% until it comes to a halt.

**Expected behavior**
The combination of jellyfin 10.6.4 and kodi 18 worked fine, no matter how many times I reset the local database or whatnot.

**Logs**
https://pastebin.com/n0e3SGFe

**Screenshots**
![bug](https://user-images.githubusercontent.com/13797/107808615-2d5d9200-6d6a-11eb-8ca5-0e4e6306c952.jpg)
(Smudge is intentional)

Jellyfin perspective:
```
[2021-02-12 19:18:21.424 +01:00] [INF] [102] Emby.Server.Implementations.ScheduledTasks.TaskManager: "Scan Media Library" Completed after 7 minute(s) and 53 seconds
[2021-02-12 19:18:21.424 +01:00] [INF] [102] Emby.Server.Implementations.ScheduledTasks.TaskManager: ExecuteQueuedTasks
[2021-02-12 19:18:21.472 +01:00] [INF] [108] Emby.Server.Implementations.IO.LibraryMonitor: Watching directory /data/videos/movies
[2021-02-12 19:18:21.477 +01:00] [INF] [106] Emby.Server.Implementations.IO.LibraryMonitor: Watching directory /data/videos/shows
[2021-02-12 19:20:04.381 +01:00] [INF] [107] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "127.0.0.1" closed
[2021-02-12 19:20:32.043 +01:00] [WRN] [111] Emby.Server.Implementations.HttpServer.WebSocketConnection: WS "127.0.0.1" error receiving data: "The remote party closed the WebSocket connection without completing the close handshake."
[2021-02-12 19:20:32.051 +01:00] [INF] [111] Emby.Server.Implementations.HttpServer.WebSocketManager: WS "127.0.0.1" closed
[...]

PR #474, opened 2021-02-22
Fix: array.array.tostring is removed in Python 3.9

oddstr13:
Potential fix for #469


mcarlton00 (approved):
This will probably be made obsolete whenever we decide to use the upstream websocket package, but seems fine to leave it in place for now

Issue #169, opened 2019-12-20
Enabling master lock using direct paths causes content to disappear


lhupitr:
Using Jellyfin for Kodi with direct paths, initial ingestion completes normally, content is available and can be played but if master lock is enabled in Kodi (even when choosing to not lock down anything) all content disappears. Normally, when using Kodi without Jellyfin, this wouldn't cause content to disappear.

 - OS: Linux OSMC
 - Jellyfin Version: 10.4.3
 - Kodi Version: 18.5
 - Addon Version: 0.40

TrueTechy:
I'm using Ubuntu 18 so there's a difference there but otherwise it's the same and the content does not disappear when enabling the master lock. Could you try disabling the lock and seeing if the content returns?

lhupitr:
Yes, I should have mentioned, disabling the lock causes the content to return. The same thing also happens on an x86-64 pc with arch linux.

TrueTechy:
Do you have the addon set to Native mode? 

Seem this might be related to an issue with Kodi with how content is added by the addon, but this only seems to affect Native mode
https://forum.kodi.tv/showthread.php?tid=208996

lhupitr:
Yes, this occurs in native (direct paths) mode, this does not happen when using the default playback mode

TrueTechy:
We'd need to dig into the differences between how the addon adds to Kodi in the different modes. Do you have a specific need for native mode just now?

lhupitr:
Yes there are various kodi in play addons that I use.

mcarlton00:
We've received reports via Matrix that this is happening with multiple profiles in addon mode as well.

MagnificentMagician:
Same issue here with addon-mode enabled. This bug really annoys me, as having a master lock is mandatory for enforcing age restrictions on movies. The system is an OSMC Vero 4K. I've deleted the .kodi folder, tried the stable OSMC Kodi Leia & beta OSMC Kodi Matrix. After setting a master lock the libraries are empty. There are approx. hundred lines logspam of the following after updating libraries in Jellyfin manually (with master lock enabled):

2021-05-17 19:31:13.090 T:2797     INFO <general>: JELLYFIN.helper.utils -> DEBUG::jellyfin_kodi/helper/utils.py:48 --[ window clear: Jellyfin.nodes.recentepisodes.path ]
[...]

PR #482, opened 2021-03-09
Stop editing sources.xml

mcarlton00:
Best I can tell, there's no reason for this to happen.  Sync and playback still works fine and it's basically just creating empty sources right now.

### DIFF
--- a/jellyfin_kodi/full_sync.py
+++ b/jellyfin_kodi/full_sync.py
@@ -95,8 +95,6 @@ class FullSync(object):
         else:
             self.mapping()
 
-        xmls.sources()
-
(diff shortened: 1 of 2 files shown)


TrueTechy:
Is this code meant to fix the issue reported in #169 but Kodi has maybe changed and thus it's no longer working?

mcarlton00:
I don't have a clue about #169.  More native mode shenanigans.  This is mostly because there's no reason for it to exist.  It's creating an empty `smb://` and `http://` source that doesn't go anywhere, and the remote repo bit has never worked since we forked.  It seems like that bit is meant as an alternative manual update method instead of using the addon repo.

TrueTechy (approved):

PR #488, opened 2021-03-14
Fix formatting on multiline if statement

mcarlton00:
Please the linter gods


oddstr13 (approved):
I think it would've been best to split all the conditions into separate lines (so that the first `and not` is at the start of the line)

PR #500, opened 2021-04-19
Speed up sync process

mammo0:
Hello,

lately I had the problem that the "fast sync" wasn't fast at all. I had 300+ changes in my Jellyfin library. It took Kodi about two seconds (sometimes more) per item to update/add it to the Kodi database.

**So the fast sync process of 300+ items took longer than full sync with 2000+ items...**
My thought was that there must be something that slows down the fast sync process. I looked into the code and found two things:

1. During fast sync for every updated item of the Jellyfin library e.g. a `Movies` object is created. See `jellyfin_kodi/library.py`. So I changed this behavior to create the those database objects before the actual processing of the items happens. This speeded up the whole sync process at least a little bit. But the fast sync was still slower compared to a full sync...

2. Actually I stumbled about the second solution by accident. After I commented out some lines in the `movie` method (`jellyfin_kodi/objects/movies.py`) the fast sync now was incredibly fast 👍 
Ok what happend? I tracked it down to the `get_person` method (`jellyfin_kodi/objects/kodi/kodi.py`). To avoid a lot of queries to the Kodi database there was a cache for peoples created in this file. This cache is a dictionary that has the people names as keys and their unique IDs of the Kodi database as values. When querying for a person this cache should be used first, afterwards the Kodi database to get the person ID.
But the problem was that this cache was never used. When looking into the cache the `args` parameter was used as a list, because of the `*`. So the cache was always searched for a list, but it contains only strings as keys. Because of that the person IDs were fetched from the Kodi database with a `SELECT` call. And this slowed down the whole sync process.

A positive side effect of my fix: Not only the fast sync is now really fast, but also the full sync is faster.

### DIFF
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -598,6 +598,18 @@ class UpdateWorker(threading.Thread):
 
     def run(self):
         with self.lock, self.database as kodidb, Database('jellyfin') as jellyfindb:
+            default_args = (self.server, jellyfindb, kodidb, self.direct_path)
+            if kodidb.db_file == "video":
+                movies = Movies(*default_args)
+                tvshows = TVShows(*default_args)
+                musicvideos = MusicVideos(*default_args)
+            elif kodidb.db_file == "music":
+                music = Music(*default_args)
+            else:
+                # this should not happen
+                LOG.error('"{}" is not a valid Kodi library type.'.format(kodidb.db_file))
+                return
+
             while True:
 
                 try:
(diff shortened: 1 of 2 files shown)


mcarlton00 (approved):
Good call on the initializing objects only one time.  That's something that has crossed my mind before but I kept forgetting to go back and test out.

I wonder if the `get_person()` was broken since it was implemented, or if I broke it during my last bug hunt.  Either way, I'll accept the blame for that one, good catch.  Haven't tested a fast sync with a large number of items, but this saves almost 10 seconds on my 1000 movie library from a fresh sync.

Issue #265, opened 2020-04-06
Wrong mapping audio when transcoding


Ashyni:
**Describe the bug**
Hello,
When choosing the audio/sub tracks for transcoding, only the subtitle is good, the audio track seems to always be the last no matter the choice made.
Everything works correctly on the web client.
Movie mapping https://pastebin.com/hfkLQ54k .

**To Reproduce**
. Reduce video quality to force transcode
. Enable audio/subtitles selection
. Play a movie with multi audio/sub
. Select Audio track
. Select Sub track
. Watch the movie

**Expected behavior**
Watch the movie with the correct audio/sub track when using transcoding.

**Logs**
https://pastebin.com/GynM97ix

**System (please complete the following information):**
 - OS: [Android, Windows]
 - Jellyfin Version: [10.5.3]
 - Kodi Version: [18.6]
 - Addon Version: [0.5.3]

mcarlton00:
Confirmed that I'm seeing this on multi audio videos when selecting the "Jellyfin transcode" option in the context menu.

Ashyni:
Let me know if you need testing.

mcarlton00:
Just to be clear, you've set `Enable audio/subtitles selection` to "Enabled" in `Addons -> Jellyfin -> Settings -> Playback`, correct?  Fun fact, I didn't know that option existed until I started digging through the code.  Anyway, if I flip that to enabled I get a dialog when I start transcoding a file and it seems to behave properly for audio.  The file I tested had 2 tracks, one English and one Japanese, and both played correctly after choosing them

Ashyni:
Yes, i have enable this to select the track.
I was wondering if it could be a good idea to have a `Video quality selection` when using `Context menu > Jellyfin transcode` for occasional transcode, the same way audio/sub selection are.
Independent to `Settings > Playback > Video quality` and/or may be move this setting to the `Context menu > Quality transcode` for faster access/adjustment, i don't know, just an idea.

mcarlton00:
That sounds fairly involved to get working, and especially that Kodi will nativly play just about everything it has fairly minimal benefit.  It also deserves it's own issue if it's going to be discussed further.

I can't replicate your issue of the wrong audio file mapping during transcoding.  Is it still present for you?

Ashyni:
Yes, problably fix with this issue i had (CORS issue on my end https://github.com/iwalton3/jellyfin-mpv-shim/issues/77)
[...]

PR #504, opened 2021-04-26
Fix transcoding of wrong audio channel

venomousZealot:
It's a zero based index, so if the first entry is not the default language, you can't ever select it.
Probably related to https://github.com/jellyfin/jellyfin-kodi/issues/265

PS: It's my first time messing with code on github. Sorry, if I didn't follow protocol correctly.


mcarlton00 (approved):
Seems to work fine from what I can tell.  Thanks.

Issue #435, opened 2020-11-29
"Force Transcode" limits available of subtitle streams


duckinator:
**Describe the bug**

If "Force Transcode" is enabled and the prompt to choose subtitles is enabled, when a file is selected to be played it first shows the "Choose the audio stream" screen. After choosing an audio stream, it shows the "Choose the subtitles stream" screen.

The "Choose the subtitles stream" screen is not presenting all of the options. On the video I'm having problems with, there are multiple options for each language, but only one is shown. Unfortunately, the one it shows is for the "behind the scenes" audio track, not the one I want to listen to.

**To Reproduce**

1. Enable the option to choose audio track/subtitles stream. (I forget where this setting is, sorry.)
2. Enable 'Force Transcoding'
3. Play a video with multiple distinct subtitle streams for a single language.
4. See that only one stream for each language is available.

**Expected behavior**

I'd expect to be able to select all of the available subtitle streams for a given language.

**Logs**

**Screenshots**

**System (please complete the following information):**
 - OS: LibreELEC 9.2.6 (RPi4.arm)
 - Jellyfin Version: 10.6.4
 - Kodi Version: 18.9.0
 - Addon Version: 0.6.5+py2
 - Playback Mode: Add-On

**Additional context**

Jellyfin server and LibreELEC/Kodi client are on the same network. LibreELEC/Kodi is running on a Raspberry Pi 4.

The "Force Transcode" option was enabled due to an issue that appears to be the same as (or closely related to) #321.

Electry:
Are we talking about internal or external subs? Also, do you have "Allow subtitle extraction on the fly" enabled in the server dashboard?

I use transcoding almost exclusively and haven't noticed any subs to be missing.

duckinator:
- What do you mean by "internal or external subs"?
- "Allow subtitle extraction on the fly" is enabled.
- It specifically only occurs on things that have multiple subtitles for the same language. E.g., one has English subtitles + English subtitles with commentary.

mcarlton00:
> Internal vs External

Internal subs are subtitles that are embedded inside of the video file.

External subs are subtitles that are in a separate file - `My Super Awesome Movie.eng.srt` for example

I would guess that it's only showing one or the other, or possibly it's just seeing 2 subtitle tracks of the same language and only displaying one.

venomousZealot:
This is because

1. the available subtitle streams are stored in a dictionary

> /jellyfin_kodi/helper/playutils.py line 586:

[...]

PR #508, opened 2021-04-28
Use jellyfin's DisplayTitle for audio and subtitle streams

venomousZealot:
This makes the titles of audio and subtitle streams consistent with Jellyfin's web-ui. It's also more descriptive.

In addition, it partially works around this bug: https://github.com/jellyfin/jellyfin-kodi/issues/435
To fully fix this bug - that is allowing multiple streams with identical properties to show up in the list - I would suggest, replacing ordered dictionaries with something better suited. If you agree, I might look into that.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -557,23 +557,6 @@ class PlayUtils(object):
 
         return path
 
-    def get_commercial_codec_name(self, codec, profile):
-        NAMES = {
-            'ac3': 'Dolby Digital',
-            'eac3': 'Dolby Digital+',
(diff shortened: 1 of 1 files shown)


mcarlton00 (approved):
Works fine, and is less code, so you get bonus points.

mcarlton00:
> replacing ordered dictionaries with something better suited. If you agree, I might look into that.

This sounds fine.  I wonder if just appending the index of the stream from JF in the key would also be sufficient.  Either way, if you have ideas then go ahead and try.  Transcoding doesn't get used in our systems very often so it can be a bit neglected at times.  Any improvements you can think of would be appreciated.

PR #529, opened 2021-05-17
Remove double slash when transcoding with a baseurl

### LINKED ISSUE #527: extra slash when custom subdirectory is used, can't transcode.
**Describe the bug**
Jellyfin for kodi inserts an extra slash to server URL when trying to access video when transcoding. This leads to fail to reproduce videos when transcoding, but not when using direct stream. The problem only occurs if "base URL" parameter is defined (necessary when routing though nginx with subdirectory). An extra slash "/" is added after the server URL, causing a double slash in URL (i.e: http://192.168.1.100/jellyfin//video.............)

**To Reproduce**

1. In Jellyfin server, go to Advanced --> Networking, and set a subdirectory in "base URL" (in my case: "/jellyfin")
2. Connect to server using Jellyfin for Kodi. Server URL will be "http://192.168.1.100:8096/jellyfin"
3. Direct play of videos work, but if you force transcode, it will not play.

**Expected behavior**
Videos played using transcoding should play :)

**Logs**
```
2021-05-16 23:32:32.069 T:2273362800   DEBUG: Error, could not open file http://192.168.1.100:8096/jellyfin//videos/fea83c6b-ebc0-7f51-9c34-c4cf2245618c/master.m3u8?DeviceId=Unknown%20Device%20id&MediaSourceId=fea83c6bebc07f519c34c4cf2245618c&VideoCodec=h264,hevc,h265,mpeg4,mpeg2video,vc1&AudioCodec=aac,mp3,ac3,opus,flac,vorbis&PlaySessionId=[REDACTED]&api_key=[REDACTED]&TranscodingMaxAudioChannels=6&RequireAvc=false&Tag=cd2a4542fb539aa5a6e7fe842a2ba906&h264-videobitdepth=8&TranscodeReasons=ContainerNotSupported,VideoCodecNotSupported,AudioCodecNotSupported&AudioStreamIndex=1&VideoBitrate=999744000&AudioBitrate=256000&maxWidth=1920&maxHeight=1080
```

**Screenshots**
This is the setting that causes the problem:

baseurl

 - OS: LibreELEC 9.2.6 and 10.0b2
 - Jellyfin Version: 10.7.x (.0 to .5 at least)
 - Kodi Version: 18.9 and 19
 - Addon Version: At least 0.7.1-onwards
 - Playback Mode: Add-On

**Additional context**
When no baseurl is defined, transcoding works, because videos are accesed through "http://192.168.1.100/videos/......" but if "/jellyfin" is added as baseurl, the server url becomes "http://192.168.1.100/jellyfin//videos/......" because an extra slash is added at the end of the baseurl.

mcarlton00:
Fixes #527

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -247,7 +247,7 @@ class PlayUtils(object):
 
             video_type = 'live' if source['Protocol'] == 'LiveTV' else 'master'
             base = base.replace('stream' if 'stream' in base else 'master', video_type, 1)
-            self.info['Path'] = "%s/%s?%s" % (self.info['ServerAddress'], base, params)
+            self.info['Path'] = "%s%s?%s" % (self.info['ServerAddress'], base, params)
             self.info['Path'] += "&maxWidth=%s&maxHeight=%s" % (self.get_resolution())
         else:
             self.info['Path'] = "%s/%s" % (self.info['ServerAddress'], source['TranscodingUrl'])


oddstr13 (approved):
Is it guaranteed that the url components contain the `/`?
An alternate way of doing this would be `.lstrip('/')` and `.rstrip('/')` of the input.

mcarlton00:
It will be until somebody changes the server on us again.  The server address doesn't have it, but the `base` variable does.  It's populated by pulling the `TranscodingUrl` from the media source, which does have a slash.

```python
(Pdb) pp source.get('TranscodingUrl')
'/videos/56c72484-db93-27f3-2268-a462ec451662/stream.m3u8?DeviceId=Unknown%20Device%20id&MediaSourceId=56c72484db.......'
```

PR #530, opened 2021-05-17
Fix impossibility of choosing the first subtitle

venomousZealot:
The usual combination of a zero-based index and a lazy if...

Also, there is no reason to use the default subtitles after asking users which subtitles they want.


mcarlton00 (approved):
Makes sense to me

PR #546, opened 2021-07-05
Add argument for dev mode to build script

mcarlton00:
Gives an option to not exclude the git repositories.  It's nicer to have all the bits included if you're using the build script to make debug builds.

### DIFF
--- a/build.py
+++ b/build.py
@@ -73,7 +73,7 @@ def zip_files(py_version: str, source: str, target: str, dev: bool) -> None:
         for root, dirs, files in os.walk(args.source):
             for filename in filter(file_filter, files):
                 file_path = os.path.join(root, filename)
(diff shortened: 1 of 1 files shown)


build.py
```diff
@@ -73,7 +73,7 @@ def zip_files(py_version: str, source: str, target: str) -> None:
         for root, dirs, files in os.walk(args.source):
             for filename in filter(file_filter, files):
                 file_path = os.path.join(root, filename)
-                if folder_filter(file_path):
+                if folder_filter(file_path) or dev:
```
oddstr13, line 76:
```suggestion
                if dev or folder_filter(file_path):
```

crobibero (approved):

oddstr13 (approved):

PR #547, opened 2021-07-08
Fix external sub conflict

### LINKED ISSUE #541: External SRT subtitles are only shown on the first playback of a session
**Describe the bug**

External SRT subtitles are only shown on the first playback of a session. When viewing the second video, external subtitles are detected, but selecting them does not make them visible.  After restarting Kodi, the subtitle works again (for the first playback)

**To Reproduce**
1. Play a video in the movie menu and select an external SRT in the subtitles menu
2. Select a different video and select an external SRT in the subtitles menu

**Expected behavior**

The subtitle should be visible in the second video

**Logs**

I see multiple `ERROR <general>: CVideoPlayerSubtitle::OpenStream - Unable to create subtitle parser` but no further detail

 - OS: Andoird TV 9
 - Jellyfin Version:  10.7.6
 - Kodi Version: Kodi 19.1 (Also tested on Kodi 19.1@Windows 10)
 - Addon Version:  0.7.4
 - Playback Mode: Add-On

mcarlton00:
Fixes #541.  Removes the ability for a race condition by naming the downloaded sub files after the item ID instead of all sharing the same name.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -513,7 +513,7 @@ class PlayUtils(object):
                 LOG.info("[ subtitles/%s ] %s", index, url)
 
                 if 'Language' in stream:
-                    filename = "Stream.%s.%s" % (stream['Language'], stream['Codec'])
+                    filename = "%s.%s.%s" % (source['Id'], stream['Language'], stream['Codec'])
 
                     try:
                         subs.append(self.download_external_subs(url, filename))
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -426,7 +426,9 @@ class Player(xbmc.Player):
                 dirs, files = xbmcvfs.listdir(path)
 
                 for file in files:
-                    xbmcvfs.delete(os.path.join(path, file))
+                    # Only delete the cached files for the previous play session
+                    if item['Id'] in file:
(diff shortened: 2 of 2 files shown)


crobibero (approved):

YouKnowBlom (approved):

oddstr13 (approved):

Sjoerdfc:
Hi, I hope it's okay to reply in an old and merged PR, I don't know if these are still read by somebody. But I seem to have a problem related to this.

This PR is about the filename of the external subtitles downloaded through Kodi, correct? Because all the subtitles I download are named "stream.en.srt". I thought this PR should fix this?

Because all the subtitles are named the same, they are not saved. If I play one file, and download a subtitle, then play another file and download a subtitle, the previous subtitle gets overwritten. So I can't ever save them, I have to download them each time I play a file.

Sjoerdfc:
Ah, thanks, I misunderstood the issue and PR then.

But the name of the subtitle comes from the name of the mediafile. Apparently Jellyfin for Kodi passes "stream" as the filename to Kodi.

mcarlton00:
That's correct.  The URL for playing from the API is `{server_url}/Videos/{item_id}/stream`, so Kodi see's "stream" as the filename.  The code for constructing the URL is here: https://github.com/jellyfin/jellyfin-kodi/blob/master/jellyfin_kodi/helper/playutils.py#L285-L290

PR #555, opened 2021-08-13
Add transcode dialog skip option: "Media default"

jessbo:
Here's a new option for skipping the transcode dialog: Use media default.

My rationale for this option is that Jellyfin suggests a track via Default[...]StreamIndex.  When the user plays a file, those tracks will be chosen by default going forward.  In my case, previewing some alternate tracks via the web interface caused the "wrong" track to play afterwards.  This setting will always pick the default according to the media, rather than previous user-selection.

I've tried not to rock the "code style" boat _too_ much, but did add the transcode options as an enum.  The enums should arguably be uppercase, but at that point they're just globals wrapped in a useless import.

All thoughts, suggestions, testing, and reasons as to why cats are superior to dogs welcome.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -3,6 +3,7 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 
 #################################################################################################
 
+import enum
 import os
 from uuid import uuid4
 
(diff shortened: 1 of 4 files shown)


mcarlton00 (approved):
Seems like a good change to me.  One could argue about the enum being a lookup dict like some of the existing code, but imo the enum is cleaner so we'll roll with it

uebergabe:
This caused the addon to crash for me on a Leia CoreElec install. Kodi was using Python 2.7 and enum34 was not installed by default on this machine. Manually installed the enum34 package to get back online, perhaps it can be added as a pre-requesite for Python<3.4?

Issue #561, opened 2021-08-27
Consider updating the translate path method as recommened in Kodi logs.


develroo:
This is applicable for Kodi Matrix 19.1+ The method is  deprecated 

```
2021-08-27 09:52:39.034 T:2861     INFO <general>: JELLYFIN.monitor -> INFO::jellyfin_kodi/monitor.py:42 -->[ kodi scan/video ]
2021-08-27 09:52:39.041 T:2867  WARNING <general>: xbmc.translatePath is deprecated and might be removed in future kodi versions. Please use xbmcvfs.translatePath instead.
2021-08-27 09:52:39.042 T:2861     INFO <general>: JELLYFIN.monitor -> INFO::jellyfin_kodi/monitor.py:45 --<[ kodi scan/video ]
2021-08-27 09:52:39.051 T:2867  WARNING <general>: xbmc.translatePath is deprecated and might be removed in future kodi versions. Please use xbmcvfs.translatePath instead.
2021-08-27 09:52:39.059 T:2867     INFO <general>: JELLYFIN.database -> INFO::jellyfin_kodi/database/__init__.py:153 Database locked in: /home/osmc/.kodi/userdata/Database/MyVideos119.db
2021-08-27 09:52:39.135 T:2867  WARNING <general>: xbmc.translatePath is deprecated and might be removed in future kodi versions. Please use xbmcvfs.translatePath instead.
2021-08-27 09:52:39.162 T:2867     INFO <general>: Skipped 1 duplicate messages..
2021-08-27 09:52:39.162 T:2867     INFO <general>: JELLYFIN.database -> INFO::jellyfin_kodi/database/__init__.py:153 Database locked in: /home/osmc/.kodi/userdata/Database/MyMusic82.db
2021-08-27 09:52:39.454 T:2867  WARNING <general>: xbmc.translatePath is deprecated and might be removed in future kodi versions. Please use xbmcvfs.translatePath instead.

```

mcarlton00:
I'd updated this at one point, but it broke backwards compat with one of the other versions that we were supporting at the time.  I think it was Kodi 17, so we should be able to make the change now, but it will need to be verified first.

develroo:
Sure. It is fairly easy to test between versions, with the biggest change being from 18 to 19 and python2 being deprecated. Thanks for looking into it.

mcarlton00:
Repeating here so all the relevant info is in the same place:

[...]

PR #567, opened 2021-09-03
Switch xbmc.translatePath to xbmcvfs.translatePath

jessbo:
Fix the deprecation warnings mentioned in #561.

### DIFF
--- a/context.py
+++ b/context.py
@@ -6,12 +6,12 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 import os
 import sys
 
-from kodi_six import xbmc, xbmcaddon
+from kodi_six import xbmcaddon, xbmcvfs
 
 #################################################################################################
 
 __addon__ = xbmcaddon.Addon(id='plugin.video.jellyfin')
-__base__ = xbmc.translatePath(os.path.join(__addon__.getAddonInfo('path'), 'jellyfin_kodi'))
+__base__ = xbmcvfs.translatePath(os.path.join(__addon__.getAddonInfo('path'), 'jellyfin_kodi'))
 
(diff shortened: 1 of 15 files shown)


mcarlton00:
Thanks for taking a stab at this.  Unfortunately, I was wrong in my previous comments in the issue and changing this breaks Kodi 18 compatibility, which we're not planning on dropping support for until the next major Kodi release.  So it's not quite this easy.

```
2021-09-04 20:52:16.669 T:139898518480448   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <type 'exceptions.AttributeError'>
                                            Error Contents: Module <module 'xbmcvfs' (built-in)> does not have attribute "translatePath"!
                                              File "/home/matt/.kodi/addons/plugin.video.jellyfin/service.py", line 15, in <module>
                                                __base__ = xbmcvfs.translatePath(os.path.join(__addon__.getAddonInfo('path'), 'jellyfin_kodi'))
                                              File "/home/matt/.kodi/addons/script.module.kodi-six/libs/kodi_six/utils.py", line 110, in __getattr__
```

I suspect we'll have to make a wrapper function (probably `jellyfin_kodi/helper/utils.py` makes the most sense) that checks whether it's currently running on 18 or 19 and use the correct library based on that.  If you want to keep working on this then great, otherwise let me know and we'll close this for now.

jessbo:
Thanks for taking a look at this.  I had assumed that with such a blaring warning Kodi 18 would have already handled this.  Should have tested it more thoroughly first; sorry for the extra trouble.

Since it's purely cosmetic anyway, it's not really worth the effort to address right now.

PR #598, opened 2021-11-14
Attempt to re-establishe websocket communication if the connection fails

mcarlton00:
Needed to move the post_capabilities api call into the websocket wrapper.  It needs to be sent after websocket communication is established or remote control won't work.

### DIFF
--- a/jellyfin_kodi/jellyfin/ws_client.py
+++ b/jellyfin_kodi/jellyfin/ws_client.py
@@ -5,6 +5,7 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 
 import json
 import threading
+import time
 
 from kodi_six import xbmc
 
@@ -54,16 +55,21 @@ class WSClient(threading.Thread):
         LOG.info("Websocket url: %s", wsc_url)
 
         self.wsc = websocket.WebSocketApp(wsc_url,
+                                          on_open=lambda ws: self.on_open(ws),
                                           on_message=lambda ws, message: self.on_message(ws, message),
                                           on_error=lambda ws, error: self.on_error(ws, error))
-        self.wsc.on_open = self.on_open
 
+        retry_count = 0
         while not self.stop:
 
+            time.sleep(retry_count * 5)
             self.wsc.run_forever(ping_interval=10)
 
             if not self.stop and monitor.waitForAbort(5):
                 break
+
+            if retry_count < 12:
+                retry_count += 1
 
         LOG.info("---<[ websocket ]")
 
@@ -72,6 +78,21 @@ class WSClient(threading.Thread):
 
     def on_open(self, ws):
         LOG.info("--->[ websocket ]")
+        self.client.jellyfin.post_capabilities({
+            'PlayableMediaTypes': "Audio,Video",
+            'SupportsMediaControl': True,
+            'SupportedCommands': (
+                "MoveUp,MoveDown,MoveLeft,MoveRight,Select,"
+                "Back,ToggleContextMenu,ToggleFullscreen,ToggleOsdMenu,"
+                "GoHome,PageUp,NextLetter,GoToSearch,"
+                "GoToSettings,PageDown,PreviousLetter,TakeScreenshot,"
+                "VolumeUp,VolumeDown,ToggleMute,SendString,DisplayMessage,"
+                "SetAudioStreamIndex,SetSubtitleStreamIndex,"
+                "SetRepeatMode,"
+                "Mute,Unmute,SetVolume,"
+                "Play,Playstate,PlayNext,PlayMediaSource"
+            ),
+        })
 
     def on_message(self, ws, message):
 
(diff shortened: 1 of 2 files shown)


jellyfin_kodi/jellyfin/ws_client.py
```diff
@@ -54,24 +55,44 @@ def run(self):
...
         self.wsc = websocket.WebSocketApp(wsc_url,
+                                          on_open=lambda ws: self.on_open(ws),
                                           on_message=lambda ws, message: self.on_message(ws, message),
                                           on_error=lambda ws, error: self.on_error(ws, error))
-        self.wsc.on_open = self.on_open
 
+        retry_count = 0
         while not self.stop:
 
+            time.sleep(retry_count * 5)
```
oddstr13, line 65: If I read this correctly, the connection time will increase by 5s from 0-60s on every connection (attempt), which means that long-running clients will always wait 60s before connecting.

`retry_count` should probably be reset either in `on_connect` or `on_message`.

I assume `ping_interval` means the client is sending data down the TCP pipe every 10s? That's good, because in my experience, TCP sockets have a tendency to hang forever on read if no data is attempted sent down a socket that is gone on the other end.

jellyfin_kodi/jellyfin/ws_client.py
```diff
@@ -54,24 +55,44 @@ def run(self):
         LOG.info("Websocket url: %s", wsc_url)
```
oddstr13, line 55: I reqlize this isn't really related to this PR, but I had to note it down somewhere when I spotted it;
The `server.replace` above is problematic if the websocket url contains `https?` anywhere else than in the protocol.
It would be better to `server.split(':', 1)` and join again after doing the protocol swap (if swapping protocol is even required for the websocket library)

oddstr13 (changes requested):

jellyfin_kodi/jellyfin/ws_client.py
mcarlton00: Only if the connection has failed over and over.  The first attempt, it doesn't wait at all.  The second attempt, 5 seconds, third attempt 10 seconds, etc etc etc.  So the only time it's waiting for a minute before trying again is if the connection has failed at least 12 times previously.  My thoughts were if the connection just dropped, initialize it again as soon as possible, but if the server is restarting or down for maintenance it doesn't do any good to just keep hammering every few seconds to get back in.  And obviously the count will reset whenever Kodi is restarted.

jellyfin_kodi/jellyfin/ws_client.py
mcarlton00: Hmm, it is possible.  Unlikely, but it would be possible for the replace to hit multiple instances.
```python
server = server.replace('https', "wss") if server.startswith('https') else server.replace('http', "ws")
```

And yes, it is required to change the protocol for the websockets library.  Also yes, that line is mixing single and double quotes, and is therefore disgusting.  A `.split(:)` also isn't a good solution though, because if somebody is accessing their server over IP:Port it will break.  So probably keep the replace, but change it to `server.replace('https://', 'wss://')` instead.  If _that_ ever hits multiple instances in a single URL, somebody deserves to get throat punched.

oddstr13: That's why I specified `.split(':', 1)`, which performs a single split from the left hand side

jellyfin_kodi/jellyfin/ws_client.py
oddstr13: Yes, I understood the intention, but what I am saying is that it looks like the counter is never reset, so that on any subsequent reconnects it waits *at least* 5s, as that's in an infinite loop.

It makes sense to reset the counter on a successful connection, so that the first reconnect doesn't take 5s, and the second 10s if that isn't needed.

If the Kodi client runs for a week, but the connection to the server dies briefly twice a day, you currently have a guaranteed 60s delay before it attempts to reconnect again (after whatever timeout happened at the TCP/websocket layers).

I suggest moving the retry_count to the class instance, and setting it to 0 (or -1, but make sure to skip the sleep if <0 then) after successfully connecting.

jellyfin_kodi/jellyfin/ws_client.py
mcarlton00: I don't like `.split()`.  It still leaves you with a list that you then have to pull the correct index of.  Minor, but still another step.  I targeted the replace to include the `://` instead

oddstr13 (approved):
Websocket recovers from server restart and system sleep
(paused playback does not survive sleep however, but we may need to do workarounds for Kodi there, as it's the direct http connection for the video that dies)

We still have a 5s sleep on the first reconnect attempt, but that's fine I guess. (We could do -1 for initial and after successful connection, and move the `+=` to the start of the loop, before the sleep to avoid this)

PR #604, opened 2021-12-23
Add acces token in unsecured endpoints

mksdev:
I am implementing jellyfin http proxy for monitoring and strict control of the access to the service.  My main goal is to block everything other than login page if you don't have access token in form of header/cookie/query since jellyfin seems to have open endpoints for the metadata (and other security issues with non-admin users). 
This pull requests just adds query parameter `api_key` to the artwork/backdrops endpoints.

### DIFF
--- a/jellyfin_kodi/entrypoint/default.py
+++ b/jellyfin_kodi/entrypoint/default.py
@@ -529,7 +529,7 @@ def get_fanart(item_id, path, server_id=None, api_client=None):
         xbmcvfs.mkdirs(directory)
         item = api_client.get_item(item_id)
         obj = objects.map(item, 'Artwork')
-        backdrops = api.API(item).get_all_artwork(obj)
+        backdrops = api.API(item, None, api_client.config.data['auth.token']).get_all_artwork(obj)
         tags = obj['BackdropTags']
 
         for index, backdrop in enumerate(backdrops):
--- a/jellyfin_kodi/connect.py
+++ b/jellyfin_kodi/connect.py
@@ -149,7 +149,8 @@ class Connect(object):
 
         if 'PrimaryImageTag' in self.user:
             server_address = client.auth.get_server_info(client.auth.server_id)['address']
-            window('JellyfinUserImage', api.API(self.user, server_address).get_user_artwork(self.user['Id']))
+            api_key = client.auth.jellyfin_token()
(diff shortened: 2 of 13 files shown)


jellyfin_kodi/entrypoint/default.py
```diff
@@ -529,7 +529,7 @@ def get_fanart(item_id, path, server_id=None, api_client=None):
         xbmcvfs.mkdirs(directory)
         item = api_client.get_item(item_id)
         obj = objects.map(item, 'Artwork')
-        backdrops = api.API(item).get_all_artwork(obj)
+        backdrops = api.API(item, None, api_client.config.data['auth.token']).get_all_artwork(obj)
```
mcarlton00, line 532: You don't have to add the implicit `None` here for the server.  It could be written as 
```suggestion
        backdrops = api.API(item, api_key=api_client.config.data['auth.token']).get_all_artwork(obj)
```

I'm not sure which method I like better, just pointing it out.

mcarlton00 (commented):
Nothing that's super wrong code wise.  You've switched back and forth between single and couple quotes a few times, which is a little annoying.  It's also worth noting that you haven't done a database migration, so this wouldn't change any of the existing media you have synced.  It would only affect new media synced after this change was released.  (As an aside and for future reference, you shouldn't commit directly to your master, you should make a branch and make the PR from there instead.  It helps keep your fork from getting into an unhealthy state if other work happens)

The bigger issue I have with this is that I think it's fundamentally approaching the problem backwards and actually introducing new security concerns in the process.  By including the api key in the URLs you're handing to Kodi in this way, you're implicitly trusting both Kodi and any other addons that a user has installed with that key.  A key that, if you're using an admin user and Jellyfin has write access to media, could potentially have destructive consequences if somebody nefarious gets ahold of it.

IMO, if you're really concerned with security to this level, the better route would be to rewrite all of the image handling to instead of going directly to the server, have it go through the plugin.  So all of those image urls would be `plugin://video.video.jellyfin/image/id` or something along those lines.  Then we have to write new code to handle that new argument, make the requests ourselves, and hand Kodi back an image.  I would expect this to be a non-insignificant amount of effort, involve database migrations for any existing media, and have a likely end result of being dramatically slower to populate than it is now.

Ultimately, I don't have any great suggestions for handling this.  If you want the "more secure than default, without introducing other problems" approach, I'd suggest filtering Kodi's useragent in your reverse proxy.  Right now, I'm leery of changes around this and the UX/security implications of it.

PR #629, opened 2022-04-03
Add option to disable remote control

### LINKED ISSUE #601: Add option to disable remote control
The remote control feature is slightly scary from a security perspective. People will use this addon to connect to public or shared Jellyfin servers, and Jellyfin is not the most secure piece of software to begin with ([eg](https://github.com/jellyfin/jellyfin/issues/5415)). Allowing untrusted servers to simulate button presses creates a backdoor into people's home networks

It would be fantastic if a setting could be added to disable all forms of remote control, and only use the websocket connection for notifications and library update signals

mcarlton00:
Untested, because my Kodi is still fubared and keeps imploding every time I open the addon settings page, but should work.  Probably.

### DIFF
--- a/jellyfin_kodi/jellyfin/ws_client.py
+++ b/jellyfin_kodi/jellyfin/ws_client.py
@@ -95,7 +95,10 @@ class WSClient(threading.Thread):
                 ),
             })
         else:
-            self.client.jellyfin.post_capabilities({})
+            self.client.jellyfin.post_capabilities({
+                "PlayableMediaTypes": "Audio, Video",
+                "SupportsMediaControl": False
+            })
         # Reinitialize the retry counter after successful connection
         self.retry_count = 0
 
--- a/jellyfin_kodi/jellyfin/ws_client.py
+++ b/jellyfin_kodi/jellyfin/ws_client.py
@@ -9,7 +9,7 @@ import time
 
 from kodi_six import xbmc
 
-from ..helper import LazyLogger
+from ..helper import LazyLogger, settings
 
 # If numpy is installed, the websockets library tries to use it, and then
 # kodi hard crashes for reasons I don't even want to pretend to understand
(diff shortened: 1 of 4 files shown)


jellyfin_kodi/jellyfin/ws_client.py
```diff
@@ -79,21 +79,23 @@ def on_error(self, ws, error):
...
-                "SetAudioStreamIndex,SetSubtitleStreamIndex,"
-                "SetRepeatMode,"
-                "Mute,Unmute,SetVolume,"
-                "Play,Playstate,PlayNext,PlayMediaSource"
-            ),
-        })
+        if settings('remoteControl.bool'):
+            self.client.jellyfin.post_capabilities({
+                'PlayableMediaTypes': "Audio,Video",
+                'SupportsMediaControl': True,
```
oddstr13, line 85: Probably want to still send this, with `SupportsMediaControl` False and empty `SupportedCommands`?

Or is capabilities exclusively used for remote control and not media?

After this we'd still accept and act on remote control commands, if the server is made to ignore the data sent here in some way.

mcarlton00: I tried just sending `{'SupportsMediaControl': False}` and the server laughed at me and kept allowing remote control anyway.  I can't say for certain if that's the actual results or if it was just a quirk of my kodi instance crashing every 30 seconds though.  Sending an empty payload was the only reliable way I had it stop showing up in the play to menu.

oddstr13: Well, if playback still works with this, then it's fine.

mcarlton00: I can't test any of it.  If you want to test flipping the payload around and seeing if there's any difference then cool, but I'm dev-ing blind here until I figure out what's causing this addon to hard crash Kodi on my workstation.

oddstr13 (approved):
I think this PR only partially addresses #601, as it doesn't touch the code interpreting the commands sent from the server.

In theory it would still be possible to use remote control by having the server ignore the provided capabilities. This would however require a modified server.

Issue #600, opened 2021-12-14
Add an option to force transcode yuv444 videos


EdiTurn:
YUV444 is a sampling format which cannot be decoded by hardware on most devices, just like Hi10p. And due to performance of the chip, it can't be decoded smoothly on TVs by software either. 
So the option to force transcode this format by the server could be very usefull.

mcarlton00:
Since yuv444 isn't in the list of formats that we have listed as direct play compatible, I suspect it's already being transcoded by default.

https://github.com/jellyfin/jellyfin-kodi/blob/80430559b23bf33fa7b28f3ea4fe5e043a510996/jellyfin_kodi/helper/playutils.py#L307-L308

You should also know that it's not a particularly good format for streaming.  There's some settings in the server for yuv420, but I would guess that yuv444 is going to make a lot of servers struggle to transcode with any sort of efficiency.  In the long run, you'll probably have a better experience overall by taking the time to re-encode your files to a more common format.

EdiTurn:
> Since yuv444 isn't in the list of formats that we have listed as direct play compatible, I suspect it's already being transcoded by default.

Nope, it's controlled by the codec option it belong to, like h.264 or hevc. 


My current workaround is to install JellyCon by the side, and enable all force transcode options to watch yuv444 videos. I'm using i5-4690K with integrated GPU, and it works well when I transcoding a 1080P HEVC yuv444 video.

mcarlton00:
I don't suppose you know exactly what transcode option is the one that's working for you?  Jellyfin-kodi has more options than jellycon does currently, so I would expect this to be the other way around.  The only option jellycon seems to have that jellyfin-kodi doesn't is msmpeg4

EdiTurn:
To transcoding a HEVC yuv444 video or not is controlled by the HEVC option, and an AVC yuv444 video is controlled by H.264 option. I realize that by real tests. However enabling these options makes all HEVC/AVC videos transcoded, which leads to quality loss and waste of power. 

I just enable all force transcoding options in JellyCon, and keep options disabled in Jellyfin-Kodi. Then I will open JellyCon to watch yuv444 videos, and still using Jellyfin-Kodi for others.

EdiTurn:
Close #631

PR #631, opened 2022-04-04
Add an option to transcode HEVC RExt

EdiTurn:
This PR is a partial implementation of #600, focusing on HEVC RExt (HEVC Format Range Extension).


mcarlton00 (approved):
I don't have anything to test against, but the changes look fine to me.

Issue #625, opened 2022-03-28
Unnecessary Audio transcode while only video needs transcode 


EdiTurn:
**Describe the bug**
Currently, Kodi plugin provides audio option like prefered codec, bit rate and max channels. But there is no such option to keep the original audio track. 
So If video transcode is needed, the audio is also transcoded according to the settings above, even If it's unnecessary.

mcarlton00:
I believe the server had the ability to do this added in 10.8, but we haven't implemented it in Kodi yet.

EdiTurn:
I just update to 10.8.0-beta2, it seems that the sequence of posted codecs is ignored. The server always choose the same codec of the source if it's in TranscodingProfiles. 
However, I found the FLAC tracks are soundless while video transcoding, because the container HLS MPEG-TS doesn't support it.

PR #637, opened 2022-04-19
Set HLS segment type to fMp4

EdiTurn:
fix the problem mentioned in #625


EdiTurn:
Emmm....It breaks seeking or play from the middle.

EdiTurn:
Sorry but the break seems beyond my capability.

PR #644, opened 2022-05-07
Remove option: Developer mode

jessbo:
This option is unused and has no reason to be around.


mcarlton00 (approved):
We talked about removing this years ago and apparently completely forgot about it.  You're the lucky winner who gets to do the thing

PR #661, opened 2022-07-19
CI - remove use_aliases from emoji command

mcarlton00:
version 2.0 of the emoji library removed the use_aliases keyword.

https://github.com/carpedm20/emoji/releases/tag/v2.0.0


mcarlton00:
Which if I'm reading it right, is entirely unnecessary in our instance.  It's only used if you're not using the "proper" names for the emoji and using the aliases for them instead.  Which it appears that we are using the proper names, so we don't need to add non-default language flags.

oddstr13 (approved):

PR #701, opened 2022-12-30
Fix syncing of musicvideos with production year YYYYmmdd

oddstr13:
Be more lenient towards garbage metadata
Jellyfin some times puts the full YYYYmmdd in the year field.

### DIFF
--- a/jellyfin_kodi/objects/musicvideos.py
+++ b/jellyfin_kodi/objects/musicvideos.py
@@ -79,6 +79,12 @@ class MusicVideos(KodiDb):
                 update = False
                 LOG.info("MvideoId %s missing from kodi. repairing the entry.", obj['MvideoId'])
 
+        if (obj.get('ProductionYear') or 0) > 9999:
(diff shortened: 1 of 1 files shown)


jellyfin_kodi/objects/musicvideos.py
```diff
@@ -79,6 +79,12 @@ def musicvideo(self, item, e_item):
                 update = False
                 LOG.info("MvideoId %s missing from kodi. repairing the entry.", obj['MvideoId'])
 
+        if (obj.get('ProductionYear') or 0) > 9999:
```
mcarlton00, line 82:
```suggestion
        if (obj.get('ProductionYear', 0) > 9999:
```

wouldn't it be cleaner to use the built in fallback method of `.get()` instead of the `or`?

oddstr13: That's what I did first, but obj contains `None` values, so I got an error attempting to `>` `None` and a number.

mcarlton00 (approved):

PR #707, opened 2023-02-07
fix misleading if statement

kovaacs:
`('Season')` is just a string in parentheses, not a tuple.

E.g., `'Sea' in ('Season')` would be `True`


mcarlton00 (approved):
Good catch, thanks.

PR #731, opened 2023-05-06
Fix #539 "cast not available"

### LINKED ISSUE #539: Actor information and thumbnails not displayed on video Information Pages for Jellyfin library
**Describe the bug**
Actor information (names and thumbnails) not displayed on Information Page for files in the Jellyfin library.

**To Reproduce**
1. Open the Information Page for a video in the Jellyfin library

**Expected behavior**
The Information Page for a video should include actor information and thumbnails and the ability to browse by actor

**Logs**
No immediately relevant errors but happy to provide as necessary

**Screenshots**
"Cast Not Available"
image

Actor data *is* syncing into the library:
image

**System**
 - OS: Android (2017 Nvidia Shield)
 - Jellyfin Version: 10.7.5 (running in a docker container on Synology)
 - Kodi Version: 19.1
 - Addon Version: 0.7.4
 - Playback Mode: Native

**Additional context**
I am reasonably confident I have narrowed the problem down to jellyfin-kodi:
1. The Jellyfin web interface correctly displays actor information and thumbnails
2. Files that are added to the Kodi library directly (and scraped using either TMDB or TMDB-python addons) correctly display actor information and thumbnails. Two additional observations: i) when the same file is present in the Jellyfin library and the Kodi library (manually added), the Jellyfin entry will then include the actor thumbnails scraped by Kodi, and ii) selecting an actor will bring up their films in the library, and *those* information pages will populate with actor thumbnails, even if in the Jellyfin library. However, accessing Information page directly will *not* show the actor thumbnails
3. The Actor view of the library *is* correctly populated and can be browsed (see above screenshot)
4. The issue persists across all skins

Other context:
* "Include people (slow)" is enabled in plugin settings
* Have tried various combinations of Reset Local Database, Update libraries, Repair libraries, etc.

Many thanks for the excellent work in developing this package! Having recently migrated from a pure Kodi setup I have been impressed.

GregoireDruant:
Closes #539 "cast not available" message when viewing tvshow or episode info in kodi. 
Works with add-on mode and native mode as well.

The fix only consist in setting the mediatype and scrapper for each path entry in kodi database.
Normally only parent paths have this information, but jellyfin-kodi does not set parentPathIds.
Also I tryed setting the parentPathIds but this does not fix the "cast not available" message if the parent path is not a regular folder (does not work with plugin://).


oddstr13:
Apologies, I don't have time to review this for a while.

GregoireDruant:
Sorry there is an issue with this fix, regarding the information pannel for a tvshow episode.
If you put the cursor on an episode item, long press enter and chose information, it works.
However the information panel won't show up in these cases:
 - click on the episode when the default action is show information
 - if the default action on video item is "chose", chosing information from the menu that shows up when clicking on an episode
 - on a pc, type "i" when the cursor is on an episode.

If I remove the value "metadata.local" from the scrapper column, the info panel works again but without cast information.

I know how to fix this in native mode but not sure I can fix it in add-on mode.

Anyway I will commit a proper fix for native mode, that does not break anything in add-on mode.

I will do so by setting the path trail from episode to top parent path and set the mediatype and scrapper to the top parent path only, as kodi does.

GregoireDruant:
This is the proper fix, works perfectly in native mode, but does not fix add-on mode.

GregoireDruant:
After a lot of time spent to search how to make every kodi menu work in addon mode - for tvshows, which are the only videos that was not working ok - , the only way I cloud achieve it is by creating a top path entry with strPath = "plugin://plugin.video.jellyfin/" and strContent = tvshows, scrapper = metadata.local.

I found in the commits history that it used to be done like that but it was changed by Matt in in 2019, when he added the libraryId in the paths. I could not figure out why.

I did not remove his changes but I re added the top path with the plugin address and it seems to be working fine now.

So this PR includes fixes for both the add-on mode and native mode.

GregoireDruant:
I have been using this for 3 weeks both in native mode on a nvidia shield and in add-on mode on a windows pc, it has been working without issue.

Beside @oddstr13, can anyone review this PR ?

GregoireDruant:
I just added two new bugfixes to the music video library.
These are not related to bug 539, but since I already have this PR pending I figure I might as well commit here.

- Artist link to a music video was inserted using "insert or update", passing values that did not constitute an unique key, resulting in duplicate links (and appearing like duplicate videos in db, at least with my kodi settings).
- In musicvideos.py, obj['LibraryId'] and obj['LibraryName'] was incorrectly set, when they were correctly set a few lines before.
I removed the problematic lines of code (see diff).

oddstr13:
I (and the Jellyfin project) has a strong preference for smaller, easier to review PRs, so please create another PR for the unrelated changes :slightly_smiling_face: 

I'm done with school for the summer, so I'll hopefully have more time for this project going forward

GregoireDruant:
> I (and the Jellyfin project) has a strong preference for smaller, easier to review PRs, so please create another PR for the unrelated changes 🙂
> 
> I'm done with school for the summer, so I'll hopefully have more time for this project going forward

I just reverted these commits and created PR #742

jellyfin_kodi/database/jellyfin_db.py
```diff
@@ -71,6 +71,15 @@ def get_item_by_kodi_id(self, *args):
...
             return
 
+    def get_episode_kodi_parent_path_id(self, *args):
+
+        try:
+            self.cursor.execute(QU.get_episode_kodi_parent_path_id, args)
+
+            return self.cursor.fetchone()[0]
+        except TypeError:
+            return
```
mcarlton00, line 81: I agree with the github scanner about not mixing implicit and explicit returns.  I believe in this case it's better to `return 0` (or `None` explicitly, but imo keeping the same type is generally better practice when feasible) and check the return value before attempting to use it on the calling function.

jellyfin_kodi/objects/tvshows.py
```diff
@@ -377,6 +398,12 @@ def episode_add(self, obj):
             return self.episode_add(obj)
 
         self.jellyfin_db.add_reference(*values(obj, QUEM.add_reference_episode_obj))
+
+        parentPathId = self.jellyfin_db.get_episode_kodi_parent_path_id(*values(obj, QUEM.get_episode_kodi_parent_path_id_obj))
+        if obj['PathId'] != parentPathId:
```
mcarlton00, line 403:
```suggestion
        if parentPathId and obj['PathId'] != parentPathId:
```

Along with the return change from the queries, ensure the path actually exists before trying to use it.  It _shouldn't_ be an issue currently since the implicit return is `None`, but just an extra layer of security in case of weirdness happening.  And I would say necessary if we `return 0` instead.

mcarlton00 (changes requested):

jellyfin_kodi/database/jellyfin_db.py
oddstr13: I was thinking the same, but this is the code style currently used in these files, so I don't consider it a blocker, cleaner to just fix them all in one go later some time.

oddstr13:
@mcarlton00 said in chat that this PR functionally is good to go, and I think the styling lints are fine as is, considering the fact that that's how the old code base has been doing it, and as such is present all over these files.

PR #742, opened 2023-06-13
Fix music video db update

GregoireDruant:
- Artist link to a music video was inserted using "insert or update", passing values that did not constitute an unique key, resulting in duplicate links (and appearing like duplicate videos in db, at least with my kodi settings).
- In musicvideos.py, obj['LibraryId'] and obj['LibraryName'] was incorrectly set, when they were correctly set a few lines before.
I removed the problematic lines of code (see diff).
This fixes automatic updates of the music video library, new videos were not added before this change.

### DIFF
--- a/jellyfin_kodi/objects/kodi/kodi.py
+++ b/jellyfin_kodi/objects/kodi/kodi.py
@@ -147,8 +147,8 @@ class Kodi(object):
                 bulk_updates.setdefault(sql, []).append((person_id,) + args)
 
             elif person['Type'] == 'Artist':
-                sql = QU.update_link.replace("{LinkType}", 'actor_link')
-                bulk_updates.setdefault(sql, []).append((person_id,) + args)
+                sql = QU.insert_link_if_not_exists.replace("{LinkType}", 'actor_link')
+                bulk_updates.setdefault(sql, []).append((person_id,) + args + (person_id,) + args)
 
             add_thumbnail(person_id, person, person['Type'])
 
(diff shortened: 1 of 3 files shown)


oddstr13:
I would like to apologize.

Indecision of what is the best way of solving this has struck me, and I have failed to communicate this fact with you by commenting on this PR.

I did take a look around these parts and made an alternative patch a few weeks back (see #760), but that patch still leaves duplicates if the information is *changed* Jellyfin side (old ones are not deleted).
The thing is that the same person can have multiple roles, but I don't know how (or if) Jellyfin and the metadata providers handles this.

Your approach here would clobber such information if it does exist, and I haven't managed to decide if that's a good thing or not.

I will say that I'm not a huge fan of adding more SQL queries to maintain if it can be avoided.

GregoireDruant:
>The thing is that the same person can have multiple roles, but I don't know how (or if) Jellyfin and the metadata providers handles this.
> Your approach here would clobber such information if it does exist, and I haven't managed to decide if that's a good thing or not.

Hi,
No worries :)

I do not see why my approach would mess with people having different roles.
I just insert new actor_link if this one does not exist.
I do not change any entry in the actor table, thus leaving intact any pre existing actor_link entries with other videos.

I believe I am doing what the previous developer intended to do, by the previous implementation was flawed because the query is not manipulating data that constitute a primary key

oddstr13 (approved):

Issue #725, opened 2023-04-19
Sync very slow with errors caused by dates older than the epoch year (1970)


PatientLightning:
**Describe the bug**
When syncing a tv show season or episode older than 1970, the sync process throws errors and slows to a crawl

**To Reproduce**

1. Within Jellyfin, edit the metadata of an episode or season to have a release date prior to 1970. ex: 09/21/1957
2. Open Kodi and Sync Library
3. Watch the sync progress stop on the episode or season with the pre-epoch release date
4. See log for an error

**Expected behavior**
The sync shouldn't have errors for shows older than 1970

**Logs**
Here's a relevant snippet from the log
```
 
 
info <general>: JELLYFIN.jellyfin_kodi.helper.utils -> ERROR::jellyfin_kodi\helper\utils.py:491 Item date: 1957-09-21 00:00:00+00:00 --- [Errno 22] Invalid argument
	Traceback (most recent call last):
File "jellyfin_kodi\helper\utils.py", line 484, in convert_to_local
	date = date.astimezone(tz.tzlocal())
File "U:\Users\UserMgr0\AppData\Local\Packages\XBMCFoundation.Kodi_4n2hpmxwrvr6p\LocalState\addons\script.module.dateutil\lib\dateutil\tz\_common.py", line 144, in fromutc
	return f(self, dt)
File "U:\Users\UserMgr0\AppData\Local\Packages\XBMCFoundation.Kodi_4n2hpmxwrvr6p\LocalState\addons\script.module.dateutil\lib\dateutil\tz\_common.py", line 258, in fromutc
	dt_wall = self._fromutc(dt)
File "U:\Users\UserMgr0\AppData\Local\Packages\XBMCFoundation.Kodi_4n2hpmxwrvr6p\LocalState\addons\script.module.dateutil\lib\dateutil\tz\_common.py", line 222, in _fromutc
	dtoff = dt.utcoffset()
	if self._isdst(dt):
File "U:\Users\UserMgr0\AppData\Local\Packages\XBMCFoundation.Kodi_4n2hpmxwrvr6p\LocalState\addons\script.module.dateutil\lib\dateutil\tz\tz.py", line 222, in utcoffset
File "U:\Users\UserMgr0\AppData\Local\Packages\XBMCFoundation.Kodi_4n2hpmxwrvr6p\LocalState\addons\script.module.dateutil\lib\dateutil\tz\tz.py", line 291, in _isdst
[...]

PR #756, opened 2023-07-17
Update utils.py

PatientLightning:
Hello,

removed use of astimezone because it throws an error for times before the epoch on windows platforms and replaced it with timezone and timedelta functions instead

Reference Issue
https://github.com/jellyfin/jellyfin-kodi/issues/725

### DIFF
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -9,6 +9,8 @@ import os
 import sys
 import re
 import unicodedata
+import datetime
+import time
 from uuid import uuid4
 from distutils.version import LooseVersion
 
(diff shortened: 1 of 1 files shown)


oddstr13 (changes requested):
This implementation is naive, and does not take into account that the UTC offset can change.

https://www.youtube.com/watch?v=-5wpm-gesOY

I've added tests for this now, but I haven't managed to get windows CI working (at all, nor with changing timezones for testing), so I don't really have any way of testing alternative solutions for windows.

I would very much like for this timezone calculation to happen in a standard library, and not be implemented in the addon code, as it is as good as impossible to get it right, and *keep* it right. Please see the linked YouTube video for inspiration for staying away from time code (*shudder*).

It *is* tempting to simply return the UTC timestamp for dates below epoch, but I haven't decided if that's a good idea yet.

PR #766, opened 2023-08-27
use thought when calling to vfs for esoteric media

KyleSanderson:
so I barely know python, but saw that there were massive stalls for populating the database with a very tiny media library. There's still some massive slowdown on series retrieval, but the episodes load in in about a second now from minutes previously. This results in a 7x performance improvement (time reduction) for library import overall, with about a 90x reduction on per-episode import.

### DIFF
--- a/jellyfin_kodi/objects/movies.py
+++ b/jellyfin_kodi/objects/movies.py
@@ -189,13 +189,13 @@ class Movies(KodiDb):
             obj['Path'] = obj['Path'].replace(obj['Filename'], "")
 
             '''check dvd directories and point it to ./VIDEO_TS/VIDEO_TS.IFO'''
-            if validate_dvd_dir(obj['Path'] + obj['Filename']):
+            if ('Container' not in obj or ('Container' in obj and obj['Container'] == 'dvd')) and validate_dvd_dir(obj['Path'] + obj['Filename']):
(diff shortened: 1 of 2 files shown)


oddstr13 (changes requested):
You are making a bunch of duplicate logic here, I think it would make sense to move this check into the `validate_*_dir` functions (and probably just pass the `obj` in there for now).

TODO
----
- [ ] verify if `validate_*_dir` functions are also called from elsewhere

mcarlton00:
Did some benchmarking here.  Synced my entire movies and shows libraries from a fresh install on my main rig.

Current master: 4 minutes, 23 seconds
This PR: 4 minutes, 25 seconds

That's close enough I'd chalk it up to network speed variations, but also seems to indicate this is insignificant as far as speeding up the sync.  It's possible this is only meaningful on lower power devices.  Would need more investigating to say for sure.

oddstr13:
This would in theory cut out two round-trips when adding items *in native mode*.

I'm not sure if `Container` is ever `bluray` or `dvd`, but I see that is checked elsewhere too.
Is that an API change, or did this never work?

`Container` isn't included in the toplevel item object when the media is BluRay or DVD (but an empty string in the `MediaSources` entries).

`VideoType` can have the value `Dvd` or `BluRay` however. It is `VideoFile` for mkv files.

Issue #765, opened 2023-08-27
[BUG] No URL encoding for resources - unusable with a single space.


KyleSanderson:
**Describe the bug**
The Plex Kodi Connect addon does this well and is a configurable option (I'd prefer it was smart enough, but anyway). nginx is a much more mature/effecient file server than the built-in ones for a multitude of reasons. In PKC there's the ability to escape a litany of characters for SMB usage, for whatever reason it doesn't look like Jellyfin-Kodi does that, and gives me 404s because the path has a single space in it. It's less apparent with SMB, but there needs to be escaping done for all of these "special" characters regardless.

**To Reproduce**
Specify a HTTP webserver as the local content streamer.
"Unable to find media `http://web:8080/The Kittens/E01 - Smoking Oregano.mkv`"

**Expected behavior**
The litany of characters are escaped, and media is played back.

**Logs**
N/A.

**Screenshots**

**System (please complete the following information):**
 - OS: Android
 - Jellyfin Version: [INF] [1] Main: Jellyfin version: 10.9.0
 - Kodi Version: Omega quietvoid
 - Addon Version: 0.7.10
 - Playback Mode: Native

**Additional context**
I've finally admitted to myself that no one uses Kodi network filesystems natively, and only streams over HTTP. I'm trying PKC and it's ok but Plex was down for about a minute there and I couldn't register my device, which is unacceptable. Jellyfin-ATV continues to be an unmitigated disaster, but I'm hopeful Jellyfin can at least help me with library management.

KyleSanderson:
Here's the PR that fixed this 3 years ago in PKC https://github.com/croneter/PlexKodiConnect/pull/1184

PR #767, opened 2023-08-28
escape network paths properly using python

KyleSanderson:
Fixes https://github.com/jellyfin/jellyfin-kodi/issues/765


oddstr13 (changes requested):
I agree with the sonarcloud bot here, and I *really* don't like that regex.
A simple string split would do the job more efficiently.

You should also be re-using the protocol variable, and be in that block.
`if protocol in ("http", "https", "ftp", ...)`

This patch also needs some testing/double-checking to verify that we're not double-quoting values here. (escaping already escaped paths)

oddstr13:
Superseded by #792

PR #774, opened 2023-09-11
Fix syncing of mixed TV libraries

sonofsky2010:
If the library was mixed, it can not be found since the it's ID has prefix "Mixed:".


oddstr13:
I'll eventually get around to testing these changes, but in the mean time I suggest you formulate a good reason to keep the mixed library feature if it is important for your workflow, and you can't do without https://github.com/jellyfin/jellyfin-meta/discussions/46

oddstr13 (approved):

PR #780, opened 2023-10-03
Make sure pytest actually fails CI on windows

### LINKED ISSUE #777: pytest on windows does not actually fail the CI run on errors
https://github.com/jellyfin/jellyfin-kodi/actions/runs/6269825291/job/17026823037#step:7:83

nwithan8:
For whatever reason, the error code from `coverage run` is not surfacing to the Windows runner, and therefore the CI is passing despite several unit tests failing.

Using `pytest` directly does surface the proper error code and halt CI if any tests fail, so that has been added as a step.

`coverage run` is still executed, after `pytest`, to generate the necessary information needed for `coverage xml` and `coverage report`. Because `pytest` runs first, the false-negative tests in `coverage run` will never be encountered.

This fails as expected:
image

Closes #777

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -63,9 +63,6 @@ jobs:
       - name: Generate coverage report
         run: |
           coverage xml
-
-      - name: Print coverage report
-        run: |
           coverage report
 
       - name: Upload coverage
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -56,14 +56,16 @@ jobs:
           flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics --output-file=flake8.output
           cat flake8.output
 
-      - name: Test with pytest
-        run: |
-          pytest
-
-      - name: Generate Coverage
+      - name: Run tests and generate coverage
         run: |
           coverage run
+
+      - name: Generate coverage report
+        run: |
           coverage xml
+
+      - name: Print coverage report
+        run: |
           coverage report
 
       - name: Upload coverage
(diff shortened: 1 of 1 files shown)


oddstr13:
I think it would be worth testing whether `coverage run` in a separate step properly fails the test
(are tests on windows only failing based on the exit code of the last command?)

oddstr13 (changes requested):
Looks like my suspicion about only last command's exit code being used is likely correct, see nedbat/coveragepy#1003 and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

For now, just move `coverage run` to a separate step, naming it `Run tests and collect coverage`, and rename the existing one `Generate coverage reports`.

That should fix pytest not erroring (via coverage) on Windows.
---
We likely have similar issues all over CI, but that can be fixed properly in a separate PR, preferably by having multi-line test steps error out properly.

oddstr13:
It is also possible, but not guaranteed that explicitly setting `shell: bash` on the run step(s) fixes the issue.

nwithan8:
> It is also possible, but not guaranteed that explicitly setting `shell: bash` on the run step(s) fixes the issue.

I don't believe that would work, trying to run a Bash shell on a Windows machine.

nwithan8:
> Looks like my suspicion about only last command's exit code being used is likely correct, see [nedbat/coveragepy#1003](https://github.com/nedbat/coveragepy/issues/1003) and https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference
> 
> For now, just move `coverage run` to a separate step, naming it `Run tests and collect coverage`, and rename the existing one `Generate coverage reports`.
> 
> That should fix pytest not erroring (via coverage) on Windows.
> 
> We likely have similar issues all over CI, but that can be fixed properly in a separate PR, preferably by having multi-line test steps error out properly.

Better Google-fu than me!

oddstr13 (approved):

PR #781, opened 2023-10-03
Fix changelog parsing

### LINKED ISSUE #770: Reformat changelog does not support unnamed section
```python
  File "/home/runner/work/jellyfin-kodi/jellyfin-kodi/.github/tools/reformat_changelog.py", line 87, in <module>
    reformat(args.format, args.emoji)
  File "/home/runner/work/jellyfin-kodi/jellyfin-kodi/.github/tools/reformat_changelog.py", line 49, in reformat
    section["items"].append(gd)
KeyError: 'items'
```

https://github.com/jellyfin/jellyfin-kodi/actions/runs/6033384710/job/16369981650#step:5:56

https://github.com/jellyfin/jellyfin-kodi/blob/68c49ea4b7c3b57487dcb56f8564024bcb4e5926/.github/tools/reformat_changelog.py#L33-L49

nwithan8:
Fixed and commented changelog parsing for CI.

Initial comment in skipped "big header" section was causing a key error issue.

Closes #770

### DIFF
--- a/.github/tools/reformat_changelog.py
+++ b/.github/tools/reformat_changelog.py
@@ -30,24 +30,44 @@ def reformat(item_format: str, output_emoji: bool) -> None:
 
     sections = []
 
+    skip = False
     section: Union[SectionType, Dict] = {}
     for line in data:
+        # If the line is not a list item or a header, skip it
+        if not line[0] in ["#", "*", "-", "+"]:
+            continue
+
+        # Skip big header
         if line.startswith("## "):
-            pass
+            skip = True
+            continue
+
+        # Beginning of new section
         if line.startswith("### "):
+            skip = False
+            # Save the previous section
             if section:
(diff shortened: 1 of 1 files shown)


oddstr13 (changes requested):
If I'm reading this right, it silently drops changes that don't have proper section tags set, which is undesirable.

Changes that shouldn't be included are explicitly tagged as so (and don't show up here), and the presence of an uncategorized PR here is an oversight and should preferably be corrected before making a release.

For now, I think the best way to handle this here is to create a default initial section where the problematic PRs go.

---
I intend to simplify the release workflow (at the cost of some complexity in the gh actions), same as I recently did in [jprm](https://github.com/oddstr13/jellyfin-plugin-repository-manager), which would have the bump release PR continually created and updated (with the content created here), but that's out of scope for this particular PR.

PR #791, opened 2023-11-21
Fix uncaught exception in Player.report_playback()

pdf:
I see regular exceptions thrown in `Player.report_playback()`, as below:

```
2023-11-19 21:32:20.858 T:1220    error <general>: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                                   Error Type: <class 'RuntimeError'>
                                                   Error Contents: Kodi is not playing any media file
                                                     File "/storage/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/player.py", line 308, in onPlayBackSeek
                                                       self.report_playback()
                                                     File "/storage/.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/player.py", line 352, in report_playback
                                                       item['CurrentPosition'] = int(self.getTime())
                                                                                     ^^^^^^^^^^^^^^
                                                   RuntimeError: Kodi is not playing any media file
```

This is not entirely unsurprising, since this method is called from numerous places, and in [at least one case](https://github.com/jellyfin/jellyfin-kodi/blob/e95e8d6a88d706f096a5a15b3d53b1fe1214fb2c/jellyfin_kodi/monitor.py#L251) is called with a delay, meaning that media playback may have completed before the method is called.

### DIFF
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -326,7 +326,11 @@ class Player(xbmc.Player):
         if not report:
 
             previous = item['CurrentPosition']
-            item['CurrentPosition'] = int(self.getTime())
+
+            try:
+                item['CurrentPosition'] = int(self.getTime())
+            except Exception:  # at this point we should be playing and if not then bail out
+                return
 
             if int(item['CurrentPosition']) == 1:
                 return
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -329,7 +329,9 @@ class Player(xbmc.Player):
 
             try:
                 item['CurrentPosition'] = int(self.getTime())
-            except Exception:  # at this point we should be playing and if not then bail out
+            except Exception as e:
+                # getTime() raises RuntimeError if nothing is playing
+                LOG.debug("Failed to get playback position: %s", e)
                 return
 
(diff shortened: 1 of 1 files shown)


jellyfin_kodi/player.py
```diff
@@ -326,7 +326,11 @@ def report_playback(self, report=True):
         if not report:
 
             previous = item['CurrentPosition']
-            item['CurrentPosition'] = int(self.getTime())
+
+            try:
+                item['CurrentPosition'] = int(self.getTime())
+            except Exception:  # at this point we should be playing and if not then bail out
```
oddstr13, line 332: We should at least log something when catching generic exceptions.
```suggestion
            except Exception as e:
                # getTime() raises RuntimeException if nothing is playing
                LOG.debug("Failed to get playback position: %s", e)
```

pdf: Fair, I just copied this handling from elsewhere in this file, including the comment. I'm travelling atm so can't link it easily, but should be easy to find.

oddstr13: That is entirely fair – we've been doing lots of cleanup in this addon since forking, but there's still a lot left, including generic exception catching

oddstr13 (approved):

PR #792, opened 2023-11-29
escape uri of path in get_file_path for http(s), ftp(s), dav(s)

omelettedufromagee:
This change aims to import a working path into Kodi if the network shared path is using http, https, ftp, ftps, dav or davs protocol.
Currently, if a path contains a space, it is not replaced by %20 and becomes unreadable for Kodi.
requote_uri function ensures that the uri's special characters are properly encoded and passed safely.


oddstr13:
Could you please provide tests for this? A list of expected inputs -> outputs, both what's now which is incorrect, and what's correct would do, I can implement the tests myself. :)

omelettedufromagee:
Sure. Below some examples with https, it is also valid for protocols http, ftp, ftps, dav and davs.

Current behavior of path:
`https://myserver.local/TV_Shows/My TV Show/My.TV.Show.S01E01.mkv`
`https://myserver.local/TV_Shows/Law&Github/Law&Github.S01E01.mkv`

This results in Kodi trying to open and play "https://myserver.local/TV_Shows/My" or use a literal ampersant in the URL.

requote_uri passes the given URI through an unquote/quote cycle to ensure that it is fully and consistently quoted.

New behavior of path:
`https://myserver.local/TV_Shows/My%20TV%20Show/My.TV.Show.S01E01.mkv`
`https://myserver.local/TV_Shows/Law%26Github/Law%26Github.S01E01.mkv`

Moreover, I've been tested this for the past few months, I haven't got any issue. If anything, it strengthens the output to make sure a valid URL is generated for those protocols.
This resolves my issue of playing file through the Jellyfin URL from Kodi, it is way slower and buggier than playing a file directly from the webdav or ftp server.

omelettedufromagee:
Btw, #767 is similar so you should be able to close it aswell with this PR.

oddstr13:
> Btw, #767 is similar so you should be able to close it aswell with this PR.

Yes, I was planning to.

Could you test this with files that actually do contain escape sequences in the file names? Sounds like that might potentially cause issues.

oddstr13:
> Any update on this?

Waiting for verification that this doesn't cause new issues (I think it likely will), I don't have time to sit down and test it myself, at least out february, possibly longer.

mcarlton00:
I feel the need to point out that this is a niche use case of a niche use case.  Using native mode alone is a niche use case, because in 99% of situations it serves absolutely no purpose over the default addon mode.  Using native mode with other niche network shares that are basically just replicating addon mode with extra setup steps is an _extreme_ niche.  So much so that it's not documented this is possible.  And frankly it's not documented because it's arguably the most complicated way you could go about doing this.  The documentation about setting this up only mentions nfs, samba, and mounted folders as options, because those are the most logical choice.

We don't use native mode, and we certainly don't do over complicated networks shares just to make native mode work like addon mode.  So we'd rather not blindly merge something we haven't had a chance to test.  And neither of us have had time to set up a test environment for this, because lack of time to chase silly niche of niche things.  Your proposed solution had quality issues that you chose not to address when requested.  This solution needs more testing, and instead of helping to test you'd rather be dramatic about your niche use case and complaining it hasn't been merged yet.  It's either going to need more verification from users, or time from one of us to experiment.

oddstr13:
@KyleSanderson I bid you farewell good sir.

PR #803, opened 2023-12-28
Use xbmcvfs functions to read and write filesystem files

sy6sy2:
At the moment, this plugin is not working on Apple TV devices.
This is because on tvOS the virtual filesystem used by Kodi hides a more complex system where in reality any XML file of Kodi is compressed in the NSUSerDefaults dictionary.
In others words, that mean that on tvOS a Kodi plugin cannot directly read or write an XML file, instead, the add-on needs to use the `xbmcvfs` functions in order to let Kodi performs the translation between the virtual filesystem and the "real" filesystem that is OS/device dependent.

This PR modify the plugin so that the XML reads and writes use `xbmcvfs` functions.
This way, this PR allow this add-on to works on tvOS devices.

For more details, see discussion here: https://forum.kodi.tv/showthread.php?tid=374294&pid=3177147#pid3177147

Tested on macOS and Apple TV with Kodi 20.

### DIFF
--- a/jellyfin_kodi/helper/xmls.py
+++ b/jellyfin_kodi/helper/xmls.py
@@ -117,13 +117,18 @@ def verify_kodi_defaults():
 
         if xbmcvfs.exists(file_name):
             try:
-                tree = etree.parse(file_name)
+                with xbmcvfs.File(file_name) as f:
+                    b = f.read()
+                    tree = etree.ElementTree(etree.fromstring(b))
             except etree.ParseError:
                 LOG.error("Unable to parse `{}`".format(file_name))
                 LOG.exception("We ensured the file was OK above, something is wrong!")
+                tree = None
 
-            tree.getroot().set('order', str(17 + index))
-            tree.write(file_name)
+            if tree is not None:
+                tree.getroot().set('order', str(17 + index))
+                with xbmcvfs.File(file_name, 'w') as f:
+                    f.write(etree.tostring(tree.getroot()))
 
     playlist_path = translate_path("special://profile/playlists/video")
 


mcarlton00:
I don't have an apple device to test with, this seems like it should solve the problem, but it does have the potential to introduce another bug.

```python
f = xbmcvfs.File(file_name)
b = f.read()
f.close()
```

In the event there's an issue with `f.read()`, the program with error out and stop execution, but the file will never be closed and be hanging out there in memory somewhere.  Typically we'd address this using something like a `with open(file_name, 'r') as f:` context, but from a quick search it doesn't look like xbmcvfs supports that fully.  I'm not sure how likely that is to happen in this case, but given that we've already had trouble with these files multiple times in the past I'm not entirely sure I want to blindly trust it.  Open to opinions

sy6sy2:
Thanks for your answer.

Did you all to reproduce the issue with `f.read()`?
I'm agree with you, we need to be sure that the file is close in any case!
Maybe something like that is ok?:

```python
try:
    b = f.read()
except:
    print('Failed to read file')
finally:
    f.close()
```

As you said, the context feature is only available since Kodi 19 (or something like that... 😕 )

oddstr13:
If the content manager works properly for Kodi 19 and later, that is the correct way to go.
Kodi 18 and lower are explicitly no longer supported since last release, and we intend to actively remove Python 2 support and do some cleanup around that going forward.

sy6sy2:
So I will try the context manager version and if it works I will modify this PR 😉
Thx.

sy6sy2:
Context manager version just pushed.
Tried on tvOS without any problem

Vergil365:
will this be fixed in the next update?
unfurtunatly i dont understand how to apply the fix to the jellyfin addon :(

PR #804, opened 2024-01-04
Pass HDR library metadata to kodi to display logos

fezster:


Theetjuh:
Would love this, will try to manually implement this, thanks in advance!

Adam1986:
It's a pretty these changes didn't make it into the latest release of Jellyfin For Kodi.

mcarlton00 (approved):
I don't have any HDR content to actually test the display functionality, but the sync works and best I can tell the HDR flag has been available since Kodi 19, so we should be in the clear

PR #807, opened 2024-01-09
Add setting to force transcode Dobly Vision

fezster:
Added setting to allow force transcode of Dolby Vision files.

This PR includes the hdr metadata fix.

On Firestick 4k Max, it is useful to be able to force a transcode with tone mapping for Dolby Vision files, otherwise certain profiles / combination of profiles just show a blank screen. I've had to force a transcode to h264 to achieve this.

### DIFF
--- a/jellyfin_kodi/helper/api.py
+++ b/jellyfin_kodi/helper/api.py
@@ -73,7 +73,20 @@ class API(object):
 
         for track in tracks:
 
+            if "DvProfile" in track:
+                track['hdrtype'] = "dolbyvision"
+			
+            elif track['VideoRangeType'] in ["HDR10", "HDR10Plus"]:
+                track['hdrtype'] = "hdr10"
+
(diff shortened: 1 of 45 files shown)


mcarlton00:
I haven't had a chance to look through everything closely, but the main issue I see is that you shouldn't be modifying all of the language files.  That's going to break our translation service.  You should only edit the english files, and the other languages will be updated automatically as translations come in

oddstr13:
@mcarlton00's review comment about modifying language files has not been addressed.

Darkatek7:
@fezster fix this lol

Onepamopa:
Why was this not merged? Nobody bothered to fix a few language files?

oddstr13:
Pr abandoned by submitter. I don't believe the code was ever tested by anyone else.

Darkatek7:
I have the same issue as the submitter

Onepamopa:
So, test it?

PR #812, opened 2024-01-27
Fix websocket automatic reconnection

### LINKED ISSUE #797: Websocket does not reconnect after Jellyfin server restart
Sounds similar to https://github.com/jellyfin/jellyfin-kodi/issues/627

I got 3 LibreElec on RPI4/5 (2 with LE 11.0.3 official and 1 with nightly for RPI5)
All 3 working just fine , data is synced using the web sockets and with kodi sync after some down time.

When restarting Jellyfin server , no client is re-connecting using websocket , i can easily verify this by Jellyfin's web GUI (devices not shown as cast target).

At first i thought i need to wait few minutes , but it never reconnects , only restarting kodi or the addon makes the client re-connect the websocket.

**To Reproduce**
1. Check the clients availability on Jellyfin's web GUI
2. Restart Jellyfin server
3. Check and confirm clients not shown on Jellyfin's web GUI and kodi's log file

**Expected behavior**
Clients should re-connect after reasonable interval. (few minutes)

 - OS: LibreElec 11.0.3
 - Jellyfin Version: 10.8.13
 - Kodi Version: 20.2
 - Addon Version: 7.11.py3
 - Playback Mode: Native

smcmnz:
Changes the reconnection approach to use the reconnect function of the websocket-client library. The current approach throws an error but does not reconnect.

After reconnection a small delay was needed before the session capabilities are registered otherwise the client will not appear as user controllable.

Resolves #797

### DIFF
--- a/jellyfin_kodi/jellyfin/ws_client.py
+++ b/jellyfin_kodi/jellyfin/ws_client.py
@@ -35,7 +35,6 @@ class WSClient(threading.Thread):
 
         self.client = client
         threading.Thread.__init__(self)
-        self.retry_count = 0
 
     def send(self, message, data=""):
 
@@ -62,23 +61,18 @@ class WSClient(threading.Thread):
 
         while not self.stop:
 
-            time.sleep(self.retry_count * 5)
-            self.wsc.run_forever(ping_interval=10)
+            self.wsc.run_forever(ping_interval=10, reconnect=10)
 
             if not self.stop and monitor.waitForAbort(5):
                 break
-
-            # Wait a maximum of 60 seconds before retrying connection
-            if self.retry_count < 12:
-                self.retry_count += 1
-
-        LOG.info("---<[ websocket ]")
 
     def on_error(self, ws, error):
         LOG.error(error)
 
     def on_open(self, ws):
-        LOG.info("--->[ websocket ]")
+        LOG.info("--->[ websocket opened ]")
+        # Avoid a timing issue where the capabilities are not correctly registered
+        time.sleep(1)
         if settings('remoteControl.bool'):
             self.client.jellyfin.post_capabilities({
                 'PlayableMediaTypes': "Audio,Video",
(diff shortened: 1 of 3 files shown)


release.yaml
```diff
@@ -65,4 +65,4 @@ dependencies:
     - addon: 'script.module.addon.signals'
       version: '0.0.5+matrix.1'
     - addon: 'script.module.websocket'
-      version: '0.58.0+matrix.1'
+      version: '1.6.4'
```
oddstr13, line 68: Have you verified that this version of the websocket library actually exists in the Kodi repositories for the supported versions?

jellyfin_kodi/jellyfin/ws_client.py
```diff
@@ -62,23 +61,18 @@ def run(self):
 
         while not self.stop:
 
-            time.sleep(self.retry_count * 5)
-            self.wsc.run_forever(ping_interval=10)
+            self.wsc.run_forever(ping_interval=10, reconnect=10)
```
oddstr13, line 64: Looks like the main functional change in this PR is the addition of the reconnect parameter here.

Could you elaborate on the reasoning behind removing the back-off delay?

release.yaml
smcmnz: Yes this version is available for Matrix, Nexus and Omega

jellyfin_kodi/jellyfin/ws_client.py
smcmnz: The reconnect param makes the reconnect logic handled by the websocket library. I added logging to the backoff delay and found it wasn't getting called anymore so I removed it for clarity.

glogiotatidis:
Is there something we can do to push this fix fwd?

IosifZ:
Just discovered Jellyfin (moving away from Plex) and this little issue is driving me crazy :)
Any timeline on when this issue will be solved?

PR #829, opened 2024-02-26
Add user to session non-interactively

rubdos:
Allows to add/remove a user to/from a session through an API call.

Use case: I use presence tracking via Home Assistant, and this allows me to update the session user list via whomever I detect present in a room/home.

This is probably not the cleanest code I've ever produced... but in order to get it cleaner, I'd first need to eat something, and I got nerd-sniped. Sorry!


oddstr13:
Not sure if I understand how this is supposed to be used.
Could you provide some more details on the use case for this?

PR #831, opened 2024-02-27
Quick and dirty movie_view fix.

AmonFlorian:
Fixes movieview from Kodi 21 beta3 (videoversion implementation).
Compatible to versions w/o this feature (Kodi 20.4 tested).

Known issues:
Movies are added as "Standard Edition" no matter their real edition.

### DIFF
--- a/jellyfin_kodi/objects/kodi/movies.py
+++ b/jellyfin_kodi/objects/kodi/movies.py
@@ -48,6 +48,11 @@ class Movies(Kodi):
     def add(self, *args):
         self.cursor.execute(QU.add_movie, args)
 
+    def add_videoversion(self, *args):
+        self.cursor.execute(QU.check_video_version)
+        if self.cursor.fetchone()[0]==1 : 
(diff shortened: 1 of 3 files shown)


squidrpi:
Movies blank for me with the script changes and reset everything. Running on a RPi5 Libreelec nightly-20240302. I have no records in videoversiontype. videoversion count matches movie table.

mcarlton00:
Finally managed to track down one of my Pis and got LibreElec nightly installed onto it.  This does indeed fix it for all new stuff added to the database the fix has been applied.  However, it does require a "repair database" to fix old entries.  Therefore, we have two options:

* Detect the update, trigger a repair database on the movies library (painful for users)
* Detect the update, perform a database migration to update old entries (painful for us)

I'll need to think more on that, it'll _probably_ happen in a separate PR, but I'll leave this open for the moment in case @oddstr13 has opinions

mcarlton00 (approved):

PR #834, opened 2024-03-07
Add automatic database migration for Kodi Omega

### LINKED ISSUE #808: Movies are empty with latest Kodi
**Describe the bug**
When I install Jellyfin video plugin to Kodi and sync my library, Movies are empty. Other libraries like TV Shows are properly synchornized. Looks like video plugin "owns" Movies (I can't add folders there) but movies are not pulled.
I found this problem on Android TV, but it also happens on PC and probably on other platforms, my logs are from PC.

This is a recent regression. I checked these revisions:
https://github.com/xbmc/xbmc/commit/5198ea50825e1f5adbd55b467eb3982fa8618766 (master @ 2024-Jan-13) - bugged
https://github.com/xbmc/xbmc/commit/979dbf5fe5d0b42daf633434afc3986651c8646a (21.0b2-Omega @ 2023-Dec-9) - working.

**To Reproduce**
1. Install Kodi
2. Setup Jellyfin server, add library for Movies
3. Install Jellyfin plugin to Kodi, connect to the server, synchronize
4. See that Movies are still empty

**Expected behavior**
Movies are synchonized.

**Logs**
Failure (master) - https://paste.kodi.tv/sipakinizu.kodi
Success (21.0b2-Omega) - https://paste.kodi.tv/uhavuyogus.kodi

 - OS: [Windows, Android]
 - Jellyfin Version: 10.8.12 (241520813c26)
 - Kodi Version:  master https://github.com/xbmc/xbmc/commit/5198ea50825e1f5adbd55b467eb3982fa8618766
 - Addon Version: 0.7.11+py3
 - Playback Mode: Any

mcarlton00:
Builds off of #831 to fix #808 

Detects if the video database version is 131 or higher (Kodi Omega beta3) to trigger a migration.  Sets all existing movies that are missing a version as a "standard" version like the original PR.

Also cleans up a few linter spacing complaints from the original PR that I missed during my review.

### DIFF
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -11,6 +11,7 @@ from six.moves import queue as Queue
 from kodi_six import xbmc, xbmcgui
 
 from .objects import Movies, TVShows, MusicVideos, Music
+from .objects.kodi import Movies as KodiDb
 from .database import Database, jellyfin_db, get_sync, save_sync
 from .full_sync import FullSync
 from .views import Views
@@ -121,6 +122,14 @@ class Library(threading.Thread):
             if not db_version:
                 # Make sure we always have a version in the database
                 db.add_version((TARGET_DB_VERSION))
+
+        # Video Database Migrations
+        with Database('video') as videodb:
+            vid_db = KodiDb(videodb.cursor)
+            if vid_db.migrations():
+                LOG.info('changes detected, reloading skin')
+                xbmc.executebuiltin('UpdateLibrary(video)')
+                xbmc.executebuiltin('ReloadSkin()')
 
     @stop
     def service(self):
--- a/jellyfin_kodi/objects/kodi/movies.py
+++ b/jellyfin_kodi/objects/kodi/movies.py
@@ -50,7 +50,7 @@ class Movies(Kodi):
 
     def add_videoversion(self, *args):
         self.cursor.execute(QU.check_video_version)
-        if self.cursor.fetchone()[0]==1 : 
+        if self.cursor.fetchone()[0] == 1:
             self.cursor.execute(QU.add_video_version, args)
 
     def update(self, *args):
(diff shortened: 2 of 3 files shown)


oddstr13 (approved):
I think it would be a good idea to have a look at how Django handles database migrations.
I think it is similar to this, except migration version is tracked in a table, and there's some magic to detect modules which are migrations.

This looks good for now tho.

No complications introduced for Kodi 20 as far as I can tell, but I've yet to test migration from 20 -> 21. 

I'd like to test the migration itself, but I don't have the opportunity for a few days yet, so if you feel like yolo merging and pushing out a release now, that is fine with me.

mcarlton00:
That was my initial plan, and the purpose of the version field in that `jellyfin.db` database.  But trying to use that left me in an interesting problem where the migration would happen even if we weren't on the newest Kodi, and then when you did update Kodi it wouldn't run again.  Ultimately that would end with us needing to keep a mapping of jellyfin db version, kodi version, and database versions to know when to trigger a migration and it would be a great big pile of not fun.  Plus, I have half a plan to completely redesign the metadata scans before we have to do another one of these.

imo waiting for more testing is fine.  First migration we've done like this, afaik, so testing is good.

oddstr13:
I can't seem to find actual packages of the Kodi RCs, and I can't be bothered to figure out how to build it.

I can still watch my series in Kodi 20 with this patch locally. :man_shrugging:

quixotic120:
also missing movie library with latest RC/coreelec and 1.0.0, same behavior as 0.7.12. can post log as well once I get home from work

mcarlton00:
Logs aren't helpful in this case.  On @kontell's first log and @quixotic120 log, they both indicate the migration ran without errors.  @kontell's second log the migration never ran, because it's not an upgrade at that point, it's a fresh sync.  And also no errors that I can see.  And of course, I can't replicate the issue on my desktop or on a libreelec nightly build.

If you want to know if this is the same issue, you need to look in the `videoversions` table of the database.  If it's populated, then it's not related to this PR and is something completely different.

"Clean library" has never worked, and will never work in the addon's current design.  This is a _very_ longstanding issue because we're doing what the Kodi devs explicitly say not to do and it can't be helped.  Don't click that button.  If you need to modify Jellyfin things in the database, go through the addon.

Lastly, if you're continuing to have an issue, please go to the actual issue and don't spread information across multiple places.  _Especially_ as this PR wasn't even what actually fixed said issue and simply made a migration for the fix somebody else implemented.

PR #841, opened 2024-03-25
Use sudo for kodirepo command

joshuaboniface:


oddstr13 (approved):
Context: publish step for v1.0.0 broke here due to permission changes during repo restructure

PR #862, opened 2024-05-17
Check for KodiSyncQueue GetPluginSettings endpoint

### LINKED ISSUE #861: library update broken with last Jellyfin server update
Hello

Since I updated Jellyfin server to 10.9.0 and 10.9.1 the library won't update automatically after starting Kodi.
A manual library update within the Jellyfin addon in Kodi works however.

I'm currently running Jellyfin server on Openmediavault and Kodi 21.0.1 on a Nvidia Shield  Pro 2019 with latest Jellyfin for Kodi addon 1.0.2+py3

oddstr13:
To determine whether the plugin is enabled or not.

Does not check if any of the items to track settings are enabled, but assumes at least one of them are.

Fixes #861

### DIFF
--- a/jellyfin_kodi/jellyfin/api.py
+++ b/jellyfin_kodi/jellyfin/api.py
@@ -6,6 +6,7 @@ import json
 import requests
 from six import ensure_str
 
+from ..helper.exceptions import HTTPException
 from ..helper.utils import settings
 from ..helper import LazyLogger
 
@@ -267,6 +268,10 @@ class API(object):
         except requests.RequestException as e:
             LOG.warning("Error checking companion installed state: %s", e)
             if e.response.status_code == 404:
+                return False
+        except HTTPException as e:
+            LOG.warning("Error checking companion installed state: %s", e)
+            if e.status == 404:
                 return False
 
         return None
--- a/jellyfin_kodi/jellyfin/api.py
+++ b/jellyfin_kodi/jellyfin/api.py
@@ -254,37 +254,20 @@ class API(object):
             'ParentId': parent_id
         })
 
-    def get_plugins(self):
-        return self._get("Plugins")
-
-    def check_companion_installed(self):
+    def check_companion_enabled(self):
         """
+        True = Enabled
+        False = Not enabled
         None = Unknown
-        True = Installed, but possibly not loaded right now
-        False = Not installed, scheduled for uninstalling or disabled
         """
         try:
-            kodi_sync_queue = [
-                x
-                for x in self.get_plugins()
-                if x.get("Id") == "771e19d653854cafb35c28a0e865cf63"
-            ]
+            plugin_settings = self._get("Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings") or {}
+            return plugin_settings.get('IsEnabled')
 
-            LOG.debug("KodiSyncQueue Plugins result: %s", kodi_sync_queue)
-
-            kodi_sync_queue_filtered = [
-                x
-                for x in kodi_sync_queue
-                if x.get("Status")
-                in ["Active", "Restart", "Malfunctioned", "NotSupported"]
-            ]
-
-            if kodi_sync_queue_filtered:
-                return True
-            else:
-                return False
         except requests.RequestException as e:
             LOG.warning("Error checking companion installed state: %s", e)
+            if e.response.status_code == 404:
+                return False
 
         return None
 
--- a/jellyfin_kodi/helper/exceptions.py
+++ b/jellyfin_kodi/helper/exceptions.py
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
 from __future__ import division, absolute_import, print_function, unicode_literals
 
+import warnings
+
 #################################################################################################
 
 
 class HTTPException(Exception):
     # Jellyfin HTTP exception
     def __init__(self, status, message):
+        warnings.warn(f'{self.__class__.__name__} will be deprecated.', DeprecationWarning, stacklevel=2)
         self.status = status
         self.message = message
 
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -353,7 +353,7 @@ class Library(threading.Thread):
                 'kodiCompanion.bool'
             ):
                 # None == Unknown
-                if self.server.jellyfin.check_companion_installed() is not False:
+                if self.server.jellyfin.check_companion_enabled() is not False:
 
                     if not self.fast_sync():
                         dialog("ok", "{jellyfin}", translate(33128))


mcarlton00:
If the plugin _is not_ installed, this is no good.  Doesn't trigger a warning in the UI and throws an error in the logs.

```
2024-05-17 10:08:32.514 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.database -> INFO::jellyfin_kodi/database/__init__.py:154 Database locked in: /home/matt/.kodi/userdata/Database/MyMusic83.db
2024-05-17 10:08:32.522 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.objects.kodi.movies -> INFO::jellyfin_kodi/objects/kodi/movies.py:162 Starting migration for Omega database changes
2024-05-17 10:08:32.523 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.objects.kodi.movies -> INFO::jellyfin_kodi/objects/kodi/movies.py:172 Omega database migration is complete
2024-05-17 10:08:32.685 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.jellyfin.http -> ERROR::jellyfin_kodi/jellyfin/http.py:124 404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings
2024-05-17 10:08:32.687 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.library -> ERROR::jellyfin_kodi/library.py:386 (404, HTTPError('404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings'))
                                                     File "jellyfin_kodi/jellyfin/http.py", line 96, in request
                                                       r.raise_for_status()
                                                     File "/home/matt/.kodi/addons/script.module.requests/lib/requests/models.py", line 1021, in raise_for_status
                                                       raise HTTPError(http_error_msg, response=self)
                                                   requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings

                                                   During handling of the above exception, another exception occurred:

                                                     File "jellyfin_kodi/library.py", line 356, in startup
                                                       if self.server.jellyfin.check_companion_enabled() is not False:
                                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                     File "jellyfin_kodi/jellyfin/api.py", line 264, in check_companion_enabled
                                                       plugin_settings = self._get("Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings") or {}
                                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                     File "jellyfin_kodi/jellyfin/api.py", line 61, in _get
                                                       return self._http("GET", handler, {'params': params})
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                     File "jellyfin_kodi/jellyfin/api.py", line 58, in _http
                                                       return self.client.request(request)
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                     File "jellyfin_kodi/jellyfin/http.py", line 159, in request
                                                       raise HTTPException(r.status_code, error)
                                                   jellyfin_kodi.helper.exceptions.HTTPException: (404, HTTPError('404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings'))
```

Not even hitting the `except requests.RequestException as e:` block so it can't proceed and just blows up.  Even fails to establish a websocket connection after this.

-------------------------------------------------------

If the plugin _is_ installed, then things appear to work properly even without an admin user.

h3llrais3r:
Maybe a suggestion: Can't we append a token in the url, in the same way it's done for the websocket connection?
image

I've tested it via postman on my local installation, and that seems to work as well (although I've generated an apikey and used that, as I don't know where the `token` comes from for the websocket client and if it works with that as well)
image

mcarlton00:
We can add auth to this request, yes.  However that's not the problem.  The `/Plugins` endpoint is now only available to admin level users.  General users can't access it, so even if they're authenticated they the existing check would still fail.

h3llrais3r:
> Not even hitting the `except requests.RequestException as e:` block so it can't proceed and just blows up. Even fails to establish a websocket connection after this.

As mentioned in the issue, the `requests.exceptions.*` are caught in the `http.py` file and a `HTTPException` is raised instead. That's why you need to catch the `HTTPException` instead. Might be worth a try to see if it then works. (At least that did the trick in my workaround)

jellyfin_kodi/jellyfin/api.py
```diff
@@ -254,37 +254,20 @@ def get_items_theme_song(self, parent_id):
...
-            ]
-
-            if kodi_sync_queue_filtered:
-                return True
-            else:
-                return False
+            plugin_settings = self._get("Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings") or {}
+            return plugin_settings.get('IsEnabled')
+
         except requests.RequestException as e:
```
h3llrais3r, line 267: Replacing this with `except HTTPException as e:` should make sure the exception is caught.

h3llrais3r:
> We can add auth to this request, yes. However that's not the problem. The `/Plugins` endpoint is now only available to admin level users. General users can't access it, so even if they're authenticated they the existing check would still fail.

Then I think the suggestion above could fix it, as long as the proper exception is caught.

oddstr13:
> If the plugin _is not_ installed, this is no good. Doesn't trigger a warning in the UI and throws an error in the logs.
> 
> ```
> 2024-05-17 10:08:32.685 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.jellyfin.http -> ERROR::jellyfin_kodi/jellyfin/http.py:124 404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings
> 2024-05-17 10:08:32.687 T:3058698    info <general>: JELLYFIN.jellyfin_kodi.library -> ERROR::jellyfin_kodi/library.py:386 (404, HTTPError('404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings'))
>                                                    Traceback (most recent call last):
>                                                    requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings

>                                                    jellyfin_kodi.helper.exceptions.HTTPException: (404, HTTPError('404 Client Error: Not Found for url: http://{jellyfin-server}/Jellyfin.Plugin.KodiSyncQueue/GetPluginSettings'))
> ```
> 
> Not even hitting the `except requests.RequestException as e:` block so it can't proceed and just blows up. Even fails to establish a websocket connection after this.

Should've known that all the layers of requests abstraction would bite me here.
Thanks for actually testing.

Properly catching the proper exception (which I naively assumed it already was) is the solution here.

h3llrais3r (approved):
I copied the changes to one of my chromecast instances to test it.
Upon start, library is synced again, so good for me.

mcarlton00 (approved):

PR #864, opened 2024-05-20
[PEP 632] Update utils.py to use 'packaging' instead of deprecated 'distutils'

leonsbane:
[PEP 632](https://peps.python.org/pep-0632/) deprecates the distutils module.

The module was marked as deprecated in Python 3.10 and 3.11, and was eventually removed in Python 3.12.

The PEP advises to replace the usage of `distutils.version` with `packaging.version`, the `LooseVersion` functionality being provided via its `parse` method.

This replacement advice is detailed in the PEP under '[Migration Advice](https://peps.python.org/pep-0632/#migration-advice)'

### DIFF
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -10,7 +10,7 @@ import sys
 import re
 import unicodedata
 from uuid import uuid4
-from distutils.version import LooseVersion
+from packaging.version import parse
 
 from dateutil import tz, parser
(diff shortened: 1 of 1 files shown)


oddstr13:
This is dead code, only other mention of compare_version is it getting imported in the `__init__.py` file for re-export.

Please delete the function and the relevant imports instead :smiley: 

(amend and force-push your commit, or create a new PR if that is easier for you)

PR #868, opened 2024-06-02
Add tool to run black on the whole project

oddstr13:
Makes it easy to re-create the PR running black on the entiere project.

With this, merge-conflicts caused by other PRs are a non-issue.

### DIFF
--- a/.github/tools/run_black.sh
+++ b/.github/tools/run_black.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 # Exit on error, print commands
-set -ex
+set -o errexit
+set -o xtrace
+
 #
 # Copyright (c) Odd Stråbø, 2024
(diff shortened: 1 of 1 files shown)


.github/tools/run_black.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+# Exit on error, print commands
+set -ex
```
joshuaboniface, line 3: Only suggestion is  that 

```
set -o xtrace
set -o errexit
```

is clearer.

joshuaboniface (approved):

.github/tools/run_black.sh
oddstr13: Usually see the short version in system scripts, so I didn't even think of the possibility

PR #880, opened 2024-06-11
Remove dependency on six

oddstr13:
Remove all usage of six and kodi_six.

- [ ] Squash commits?
- [ ] Add commits to ignore?

A few changes have potential of *maybe* being functional, but I don't think they are in modern python/kodi versions.


oddstr13:
The internal restart/reload addon function seems to be broken, but I don't think that is caused by this PR.
Fixing it is out of scope for this PR in any case.

cvium (approved):

PR #892, opened 2024-07-22
Add AV1 as a preferred codec option

Insprill:
This adds AV1 as an option for the preferred codec in addition to H264/5, and allows it to be used as a last-resort if other formats are unavailable. This could be considered a follow-up to https://github.com/jellyfin/jellyfin-kodi/pull/545, since it only allowed AV1 to be direct-played, not transcoded.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -390,6 +390,11 @@ class PlayUtils(object):
         elif not settings("transcode_h265.bool"):
             codecs.insert(1, "hevc")
 
+        if settings("videoPreferredCodec") == "AV1":
+            codecs.insert(0, "av1")
+        elif not settings("transcode_av1.bool"):
+            codecs.append("av1")
+
         if settings("transcode_mpeg2.bool"):
             codecs.remove("mpeg2video")
 
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -54,7 +54,7 @@
 		<setting label="30524" id="transcode_vc1" type="bool" default="false" visible="true" />
 		<setting label="30525" id="transcode_vp9" type="bool" default="false" visible="true" />
 		<setting label="30526" id="transcode_av1" type="bool" default="false" visible="true" />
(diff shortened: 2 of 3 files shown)


oddstr13:
Don't mind the code coverage report, adding tests for most things in this project is difficult.

jellyfin_kodi/helper/playutils.py
```diff
@@ -399,6 +399,12 @@ def get_transcoding_video_codec(self):
         if settings("transcode_vc1.bool"):
             codecs.remove("vc1")
 
+        if settings("transcode_av1.bool"):
+            codecs.remove("av1")
+        else:
+            if settings("videoPreferredCodec") == "AV1":
```
oddstr13, line 405: This should be an `elif`.
Would be cleaner to implement this the same way as HEVC is now done.

I am undecided as how to handle the race condition of setting preferred codec while asking to transcode that same codec.

oddstr13 (changes requested):
Needs a rebase due to the HEVC PR touching the same file region.
That same PR added tests covering this function, so those tests need updating to account for the changes here.

jellyfin_kodi/helper/playutils.py
Insprill: I think the best solution would be hiding the transcode option from the menu and assuming false when it's preferred.

oddstr13: That is probably the best option, yea

oddstr13 (approved):
Looks good, I just need to find a way to test if it actually works

oddstr13:
It works? Probably, sort of?
My server was screaming bloody murder, and my Kodi client didn't wanna play the video, only audio (I'm stuck on Kodi 20).

So, this PR *works*, after enabling AV1 transcoding on the server, but it doesn't actually *work* for *me*.

I tested by setting AV1 as the preferred codec, and enabling transcode HEVC.

oddstr13:
@Insprill is this feature working as intended on your side?

We got a report saying it doesn't work at all, please see jellyfin/jellyfin#15646

Insprill:
I haven't used Kodi since I wrote this since I was never able to get it working reliably my RockPi 5B, so I honestly can't say. Sorry.

PR #918, opened 2024-09-19
Use email.utils for parsing server time

### LINKED ISSUE #916: save_last_sync try to convert HTTP date header with a localized python date-time parser which is incorrect
**Describe the bug**
When the locale has a month name format "%b" which is not English save_last sync raise an exception because it cannot parse the server-date (server-date is the HTTP header "Date" of the HTTP request, from `.kodi/addons/plugin.video.jellyfin/jellyfin_kodi/jellyfin/http.py`: `self.config.data["server-time"] = r.headers["Date"]`).

HTTP Date has three format https://datatracker.ietf.org/doc/html/rfc9110.html#section-5.6.7, the "preferred format is a fixed-length and single-zone subset of the date and time specification used by the Internet Message Format [[RFC5322](https://datatracker.ietf.org/doc/html/rfc5322)]". That is https://datatracker.ietf.org/doc/html/rfc5322#section-3.3.

strptime is wrong as HTTP Date is not localized, we should use a dedicated function.

Seems the closest to parse HTTP date is to parse it as an email date with email.utils.parsedate_to_datetime. https://stackoverflow.com/questions/1471987/how-do-i-parse-an-http-date-string-in-python (not that I don't have enough reputation to comment on the wrong answers that is rfc822 which is not more appropriate that email.utils as rfc822 is the obsolete RFC for email dates (the current one is https://datatracker.ietf.org/doc/html/rfc2822#section-3.3 which is what email.utils.parse is about), not an RFC for HTTP dates. And the strptime answer that as this bug re'port shows does not work on localized systems, i.e. with an LC_TIME not English).

Also see the discussion in https://github.com/jellyfin/jellyfin-kodi/issues/288#issuecomment-617950518
I disagree that ISO8601 is the way to go because server-date is about the HTTP date. We will have to parse the HTTP date format either way, so why not keep it the IMF format from RFC5322?

The issue is that we parse a non localized datetime string with a localized function, not that the format is wrong.

But do we agree to keep RFC5322 IMF date format?

Do we agree to use email.utils.parsedate_to_datetime (it is python 3.3 and above) or should I use email.utils.parse and then convert to datetime?

And is it fine to use email.utils.parsedate rfc2822 even if I believe it does not cope with the two obsolete HTTP date formats that the  RFC9110 says we MUST be able to read?
```
  Sunday, 06-Nov-94 08:49:37 GMT   ; obsolete RFC 850 format
  Sun Nov  6 08:49:37 1994         ; ANSI C's asctime() format
```

Maybe test cases would be good too. 
Probably a test case with a locale with non-English month names, one that 2 digit year, a non rfc822 date string. Do you see others?

Edit: wed could also include djjango parse_http_date codfe which seems straightforward:
https://github.com/django/django/blob/stable/5.1.x/django/utils/http.py#L97
```
MONTHS = "jan feb mar apr may jun jul aug sep oct nov dec".split()
__D = r"(?P<day>[0-9]{2})"
__D2 = r"(?P<day>[ 0-9][0-9])"
__M = r"(?P<mon>\w{3})"
__Y = r"(?P<year>[0-9]{4})"
__Y2 = r"(?P<year>[0-9]{2})"
__T = r"(?P<hour>[0-9]{2}):(?P<min>[0-9]{2}):(?P<sec>[0-9]{2})"
RFC1123_DATE = _lazy_re_compile(r"^\w{3}, %s %s %s %s GMT$" % (__D, __M, __Y, __T))
RFC850_DATE = _lazy_re_compile(r"^\w{6,9}, %s-%s-%s %s GMT$" % (__D, __M, __Y2, __T))
ASCTIME_DATE = _lazy_re_compile(r"^\w{3} %s %s %s %s$" % (__M, __D2, __T, __Y))

def parse_http_date(date):
    """
    Parse a date format as specified by HTTP RFC 9110 Section 5.6.7.

    The three formats allowed by the RFC are accepted, even if only the first
    one is still in widespread use.

    Return an integer expressed in seconds since the epoch, in UTC.
    """
    # email.utils.parsedate() does the job for RFC 1123 dates; unfortunately
    # RFC 9110 makes it mandatory to support RFC 850 dates too. So we roll
    # our own RFC-compliant parsing.
    for regex in RFC1123_DATE, RFC850_DATE, ASCTIME_DATE:
        m = regex.match(date)
        if m is not None:
            break
    else:
        raise ValueError("%r is not in a valid HTTP date format" % date)
    try:
        year = int(m["year"])
        if year < 100:
            current_year = datetime.now(tz=timezone.utc).year
            current_century = current_year - (current_year % 100)
            if year - (current_year % 100) > 50:
                # year that appears to be more than 50 years in the future are
                # interpreted as representing the past.
                year += current_century - 100
            else:
                year += current_century
        month = MONTHS.index(m["mon"].lower()) + 1
        day = int(m["day"])
        hour = int(m["hour"])
        min = int(m["min"])
        sec = int(m["sec"])
        result = datetime(year, month, day, hour, min, sec, tzinfo=timezone.utc)
        return int(result.timestamp())
    except Exception as exc:
        raise ValueError("%r is not a valid date" % date) from exc
```

**To Reproduce**

1. Go to Jellyfin plugin in Kodi
2. Go to your library
3. Open the context menu on the library and click on Sync
4. See error in the Kodi log .kodi/temp/kodi.log

**Expected behavior**

No error in Kodi log and probably a working last sync guard.

**Logs**
```
2024-09-16 12:40:45.749 T:740028    info <general>: JELLYFIN.jellyfin_kodi.library -> ERROR::jellyfin_kodi/library.py:536 time data '16 Sep 2024 10:40:33 GMT' does not match format '%d %b %Y %H:%M:%S GMT'
                                                     File "jellyfin_kodi/library.py", line 530, in save_last_sync
                                                       time_now = datetime.strptime(
                                                                  ^^^^^^^^^^^^^^^^^^
                                                     File "/usr/lib/python3.12/_strptime.py", line 554, in _strptime_datetime
                                                       tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                                                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                                     File "/usr/lib/python3.12/_strptime.py", line 333, in _strptime
                                                       raise ValueError("time data %r does not match format %r" %
                                                   ValueError: time data '16 Sep 2024 10:40:33 GMT' does not match format '%d %b %Y %H:%M:%S GMT'
                                                   
2024-09-16 12:40:45.753 T:740028    info <general>: JELLYFIN.jellyfin_kodi.library -> INFO::jellyfin_kodi/library.py:541 --[ sync/2024-09-16T10:38:45z ]
2024-09-16 12:40:45.756 T:740028    info <general>: JELLYFIN.jellyfin_kodi.full_sync -> INFO::jellyfin_kodi/full_sync.py:227 Full sync completed in: 0:00:59
2024-09-16 12:40:45.759 T:740028    info <general>: JELLYFIN.jellyfin_kodi.full_sync -> INFO::jellyfin_kodi/full_sync.py:703 --<[ fullsync ]
```

**Screenshots**

- OS: Debian
- Jellyfin Version: 10.9.11
- Kodi Version: 21.1.0
- Addon Version: v1.0.5+py3
- Playback Mode: Add-On

**Additional context**

I tested this that seems to work.  
```
6 import threading
    7 from datetime import datetime, timedelta
    8 from email.utils import parsedate_to_datetime
    9 import queue
   10         


 527     def save_last_sync(self):
  528 
  529         try:
  530             time_now = parsedate_to_datetime(
  531                 self.server.config.data["server-time"].split(", ", 1)[1],
  532                 "%d %b %Y %H:%M:%S GMT",
  533             ) - timedelta(minutes=2)
  534         except Exception as error:
  535 
  536             LOG.exception(error)
  537             time_now = datetime.utcnow() - timedelta(minutes=2)
  538 
  539         last_sync = time_now.strftime("%Y-%m-%dT%H:%M:%Sz")
  540         settings("LastIncrementalSync", value=last_sync)
  541         LOG.info("--[ sync/%s ]", last_sync)
  542 
```

oddstr13:
Fixes #916

### DIFF
--- a/jellyfin_kodi/jellyfin/http.py
+++ b/jellyfin_kodi/jellyfin/http.py
@@ -179,12 +179,7 @@ class HTTP(object):
 
             else:
                 try:
-                    # Prefer custom Server-Time header in ISO 8601 format
-                    # TODO: Clean up once the probability of most users having
-                    #       the updated server-side plugin is high.
-                    self.config.data["server-time"] = r.headers.get(
-                        "Server-Time", r.headers.get("Date")
-                    )
+                    self.config.data["server-time"] = r.headers.get("Date")
                     elapsed = int(r.elapsed.total_seconds() * 1000)
                     response = r.json()
                     LOG.debug("---<[ http ][%s ms]", elapsed)
--- a/jellyfin_kodi/jellyfin/http.py
+++ b/jellyfin_kodi/jellyfin/http.py
@@ -179,7 +179,12 @@ class HTTP(object):
 
             else:
                 try:
-                    self.config.data["server-time"] = r.headers["Date"]
+                    # Prefer custom Server-Time header in ISO 8601 format
+                    # TODO: Clean up once the probability of most users having
+                    #       the updated server-side plugin is high.
+                    self.config.data["server-time"] = r.headers.get(
+                        "Server-Time", r.headers.get("Date")
+                    )
                     elapsed = int(r.elapsed.total_seconds() * 1000)
                     response = r.json()
                     LOG.debug("---<[ http ][%s ms]", elapsed)
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -525,18 +525,30 @@ class Library(threading.Thread):
         return True
 
     def save_last_sync(self):
+        _raw_time = self.server.config.data["server-time"]
+        # The ISO 8601 header always end with Z
+        if _raw_time and _raw_time[-1] == "Z":
+            time_now = datetime.strptime(_raw_time, "%Y-%m-%dT%H:%M:%SZ")
+        else:
+            try:
+                # TODO: Clean up once the probability of most users having
+                #       the updated server-side plugin is high.
+                LOG.warning(
+                    "Server time not in ISO 8601 format, using fallback (update KodiSyncQueue)."
+                )
+                import email.utils
 
-        try:
-            time_now = datetime.strptime(
-                self.server.config.data["server-time"].split(", ", 1)[1],
-                "%d %b %Y %H:%M:%S GMT",
-            ) - timedelta(minutes=2)
-        except Exception as error:
+                time_now = email.utils.parsedate_to_datetime(_raw_time)
 
-            LOG.exception(error)
-            time_now = datetime.utcnow() - timedelta(minutes=2)
+            except Exception as error:
+                LOG.warning(error)
+                LOG.warning("Failed to parse server time, falling back to client time.")
+                time_now = datetime.utcnow()
 
-        last_sync = time_now.strftime("%Y-%m-%dT%H:%M:%Sz")
+        # Add some tolerance in case time is out of sync with server
+        time_now -= timedelta(minutes=2)
+
+        last_sync = time_now.strftime("%Y-%m-%dT%H:%M:%SZ")
         settings("LastIncrementalSync", value=last_sync)
         LOG.info("--[ sync/%s ]", last_sync)
 
--- a/jellyfin_kodi/library.py
+++ b/jellyfin_kodi/library.py
@@ -526,24 +526,16 @@ class Library(threading.Thread):
 
     def save_last_sync(self):
         _raw_time = self.server.config.data["server-time"]
-        # The ISO 8601 header always end with Z
-        if _raw_time and _raw_time[-1] == "Z":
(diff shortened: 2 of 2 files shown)


oddstr13:
Related PR in KodiSyncQueue: jellyfin/jellyfin-plugin-kodisyncqueue#95

mcarlton00 (approved):

PR #925, opened 2024-10-01
Remove native mode

oddstr13:
Server support is getting removed in Jellyfin 10.10
See jellyfin/jellyfin#12446 for the 10.10 removal in server and jellyfin/jellyfin-web#5098 for the 10.9 removal in web.
See also jellyfin/jellyfin#11654 for some discussion around the 10.9 removal.

I was certain there was more on the topic somewhere on GitHub (I feel like there's a discussion thread missing), but it is quite possible most if not all of it was discussed only in [chat (Matrix)](https://matrix.to/#/#jellyfinorg:matrix.org) and not properly documented on GitHub.

TODO:
- [ ] Test changes
      — not much that could go wrong here, I'm just skipping the dialog and setting the config value directly
- [ ] (Optional) Add warning dialog for existing installations using "native" mode
      — add a way to silence the warning if so
- [ ] (Optional) Remove all code related to native mode
      — probably best left for later

### DIFF
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -470,19 +470,8 @@ def has_attribute(obj, name):
 
 def set_addon_mode():
     """Setup playback mode. If native mode selected, check network credentials."""
-    value = dialog(
-        "yesno",
-        translate("playback_mode"),
-        translate(33035),
-        nolabel=translate("addon_mode"),
-        yeslabel=translate("native_mode"),
-    )
 
-    settings("useDirectPaths", value="1" if value else "0")
-
-    if value:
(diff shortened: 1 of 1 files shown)


hexoverflow:
I recently updated to jellyfin 10.10 and noticed the removal of the native mode by the server. Some movies do just not play using addon mode (can reproduce it reliably using large DoVi files) and thus addon mode is unusable for me. I hacked back native mode by editing the kodi addon and doing the path substitution there.

Instead of removing this useful feature completely, it would be better to add a client setting allowing users to set the native path inside the kodi addon itself as proposed [here](https://github.com/jellyfin/jellyfin/pull/12446#issuecomment-2288497424). There's a lot of other people complaining about addon mode in the discussion as well

PR #951, opened 2024-12-14
Use file browser dialog for local path

sundermann:
Use a file browser dialog instead of a basic plaintext dialog for the local path. Kodi returns credentials for instance in smb paths selected using a browser dialog. It's better to strip these out so they do not become visible in the list.

### DIFF
--- a/jellyfin_kodi/helper/utils.py
+++ b/jellyfin_kodi/helper/utils.py
@@ -10,7 +10,7 @@ import sys
 import re
 import unicodedata
 from uuid import uuid4
-from urllib.parse import quote_plus
+from urllib.parse import quote_plus, urlparse, urlunparse
 
 from dateutil import tz, parser
 
@@ -157,6 +157,7 @@ def dialog(dialog_type, *args, **kwargs):
         "select": d.select,
         "numeric": d.numeric,
         "multi": d.multiselect,
+        "browse": d.browse,
     }
     return types[dialog_type](*args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


jellyfin_kodi/helper/utils.py
```diff
@@ -487,6 +488,13 @@ def set_addon_mode():
     LOG.info("Add-on playback: %s", settings("useDirectPaths") == "0")
 
 
+def strip_credentials(url):
```
oddstr13, line 491: I want this function in the test suite, so I'd minimum need a list of URLs to test, both with and without credentials in them.
http(s), cifs/smb, nfs and (s)ftp come to mind, there are probably others I'm forgetting.

I can write the tests myself, but I need help gathering a diverse set of example URLs (that represent your real world application) – take what you have, but replace with [example/documentation IP addresses and domain names](https://alexwlchan.net/2018/ips-for-documentation/), and anonymize paths if needed (but keep the overall structure).

The above is a note for *anyone* using native mode, *especially* those with more esoteric setups.

sundermann:
I added some test cases. Personally I could only test nfs and smb but I don't see why the other types would not work.

tests/test_helper_utils.py
```diff
@@ -77,3 +77,23 @@ def test_values(item, keys, expected):
 )
 def test_convert_to_local(utctime, timezone, expected):
     assert convert_to_local(utctime, timezone=zoneinfo.ZoneInfo(timezone)) == expected
+
```
oddstr13, line 80:
```suggestion


```

oddstr13 (approved):

PR #968, opened 2025-01-15
Fix description formatting

theguymadmax:
Before:

B1

B2

After:

A1

A2


oddstr13 (commented):
Thanks!

Looks like Kodi is parsing the string literally, as opposed to the same way HTML is parsed. Based on the content here (`&#10;` are newline characters), this is likely new.

Would've been nice to know in which Kodi release this change happened.

The changelog is also affected, but that needs to be fixed in the script populating it, not here.

oddstr13 (approved):

PR #979, opened 2025-02-25
Add paging to API/get_items

sgroen8:
This fixes getting the error "414 URI Too Long" when sending large playlists to jellyfin-kodi.


oddstr13 (approved):
With this change, I'm able to shuffle play my whole music library via casting to Kodi.
Actually loading in the whole playlist takes forever, and Kodi isn't very happy with me – but no longer errors out with URI too long.

PR #980, opened 2025-02-26
Fix default audio stream selection

fezster:
Discussion thread here: https://forum.jellyfin.org/t-wrong-audio-track-selected-by-default-jellyfin-for-kodi?pid=52729#pid52729

The current implementation does not work on Fire TV devices (Firestick 4k Max, FireCube V3). The issue seems to stem from a race condition in player.py.

If report_playback() -> detect_audio_subs() is called first, the current audio track is defaulted and this is then stored. Later, set_audio_subs() is called and this same audio track is applied. This is what should happen and, I guess, is expected to happen.

On the FireStick, though, this never happens. The set_audio_subs() is called first - which means the DefaultAudioStreamIndex returned by Jellyfin is used. The logic is then not applied correctly and the wrong audio track is applied. Later, report_playback is called, but by then the audio track has already been set incorrectly and so it never corrects itself.

By applying the correct logic to DefaultAudioStreamIndex, this then works as well.

The correct logic, I believe, should be to subtract 1 from the index returned by Jellyfin.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected audio stream selection and reporting to ensure accurate playback behavior and synchronization with the server.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -213,9 +213,10 @@ class PlayUtils(object):
             LOG.info("--[ transcode ]")
             self.transcode(source, audio, subtitle)
 
-        self.info["AudioStreamIndex"] = self.info.get("AudioStreamIndex") or source.get(
-            "DefaultAudioStreamIndex"
+        self.info["AudioStreamIndex"] = self.info.get("AudioStreamIndex") or (
+            source.get("DefaultAudioStreamIndex", 0) - 1
         )
+
         self.info["SubtitleStreamIndex"] = self.info.get(
             "SubtitleStreamIndex"
         ) or source.get("DefaultSubtitleStreamIndex")
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -382,7 +382,7 @@ class Player(xbmc.Player):
             "IsPaused": item["Paused"],
             "IsMuted": item["Muted"],
             "PlaySessionId": item["PlaySessionId"],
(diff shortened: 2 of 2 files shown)


TheOverpassArsonist:
What is the requested change by the bot? I read through it's report and it just seems to say that the pull request is valid, it doesn't seem to request any change, but Github is saying a change was requested. Have I misunderstood something or is the bot acting up? There are earlier replies that note potential issues, but the latest May 27th reply suggests that the PR is valid so I'm not following what the requested change is.

Asking because it seems like it's been a while without any activity and I'm running into an issue that I _think_ may be related. (not on a Fire TV device, LibreElec on Pi4, but I'm getting similarly inconsistent audio streams with no obvious pattern, and this being a race condition with that exact effect seems like it'd check out as the root cause.)

oddstr13:
I'm not familiar enough with these parts to properly review it, and my spider senses are tingling saying there is something strange going on somewhere here.

I have not dedicated the required hours to get to the bottom of what's going on here, mainly because I haven't personally noticed any issues with it.

Feel free to apply these changes on top of latest release/master, and report back your findings.

Rijul-A:
Possibly also covered by #1210 if anyone interested can test

oddstr13:
I didn't find much value from coderabbit when I was testing it out here previously, but just for fun if it's still enabled, I'll try to see what it says about the new PR.

Edit: ah well, the rabbit is dead.

oddstr13:
@coderabbitai please take everything discussed so far into account, and have a look at #1210 and any related issues. Could that PR solve the issue this one is attempting to fix in a better way?

PR #990, opened 2025-04-05
Update default.py

troyboyee:
I put generator_stop and annotations in the future statement. Also, this is the first pull request I ever had, so if someone could tell me if I am doing it right.


oddstr13:
A good pull-request describe the changes made, and why they are needed.
We are no longer targeting older Python versions than 3.9, so all but the annotations future import are noop, and it is only useful together with actually implementing type annotations.

https://docs.python.org/3/library/__future__.html#module-contents

PR #993, opened 2025-04-21
Update movies.py

JeroenED:
Hi,

I created this pull request as I was suffering an issue with current kodi nightly. The problem seemed pretty easy to solve, however it is a problem on nightly and I am unsure if the problem will still be there when kodi piers will go stable.

Also, I feel the problem is pretty impactful, changing the type from a now unknown to what used to be extras (I guess) makes me feel you don't want to merge this for Omega.

Anyhow, this is what is fixing the issue for me on piers right now. It might tame the dragons before they go wild. Hope you want to merge it but no hard feeling if you take omega a priority.

Reference: https://forum.jellyfin.org/t-bug-when-using-current-kodi-nighly
Since kodi piers standard version is changed to 1.
Source from the MyVideosDatabase and https://github.com/xbmc/xbmc/blob/c0be4a8e34263cb939e453dbb4ce3430b46a458c/xbmc/video/VideoManagerTypes.h#L21

### DIFF
--- a/jellyfin_kodi/objects/kodi/movies.py
+++ b/jellyfin_kodi/objects/kodi/movies.py
@@ -158,7 +158,7 @@ class Movies(Kodi):
 
         # Sets all existing movies without a version to standard version
         for entry in self.cursor.fetchall():
-            self.add_videoversion(entry[0], entry[1], "movie", "0", 40400)
+            self.add_videoversion(entry[0], entry[1], "movie", "1", 40400)
             changes = True
 
         LOG.info("Omega database migration is complete")


oddstr13:
This change will need version gating, and a separate migration step.

We also generally don't target nightlies, so this PR will have to wait until at least beta or rc, as that generally means things are starting to stabilize.

---

Looks like the database structure might be changing more substantially, but the migration seems to also be done in Kodi itself.
```cpp
// renumber itemType to free the value 0 for nodes navigation
// former value 0 for versions becomes 1
// former value 1 for extras becomes 2
```

https://github.com/xbmc/xbmc/compare/b01f67dbe312f7f0b06d29addce2b296d12c880f...6ba797d60160797af61d5a8a9314a96c21943484#diff-3e0425ad7a3f9835442da572adda9031febda02715ecac402eeb44a00579eb49R6675

A new deep dive into Kodi internals may be needed for Piers release.

JeroenED:
As I am currently on the latest nightly myself for another (non-jellyfin related) issue, I will keep this PR updated with changes required for piers. I use the capabilities of the plugin not on a daily basis both I easily see when something goes wrong. Also, you're not targetting nightlies which makes it not required that an issue is resolved on day 1.

Is this OK for you? Or is it better to create separate PRs for further changes?

oddstr13:
Yea, feel free to keep this branch up to date.
Try to keep different changes together in their own commits however, in case we end up wanting to split into multiple PRs anyways.

jellyfin_kodi/objects/kodi/movies.py
```diff
@@ -158,7 +158,7 @@ def omega_migration(self):
 
         # Sets all existing movies without a version to standard version
         for entry in self.cursor.fetchall():
-            self.add_videoversion(entry[0], entry[1], "movie", "0", 40400)
+            self.add_videoversion(entry[0], entry[1], "movie", "1", 40400)
```
oddstr13, line 161: Notes for later:
Needs separate migration step for Piers, see earlier comment.
May or may not actually need to do this migration, as looking at the Kodi source, the migration is supposed to get applied on upgrade.

Probably need a check for db version and two different queries for adding instead.

oddstr13 (changes requested):

Tualua:
Is there any way to build and install this version without waiting for approval?

JeroenED:
@pdf Sounds indeed a better option. Will see what I can do.

@Tualua You can download the repo and replace all files in $HOME/.kodi/addons/plugin.video.jellyfin (this is the linux/libreelec path; I don't know the path for windows. I guess it's somewhere in %appdata%, but that's a wild guess)

jellyfin_kodi/objects/kodi/movies.py
```diff
@@ -163,3 +166,20 @@ def omega_migration(self):
...
+        Adds a video version for all existing movies
+        """
+        LOG.info("Starting migration for Piers database changes")
+        # Tracks if this migration made any changes
+        changes = False
+        self.cursor.execute(QU.get_missing_versions)
+
+        # Sets all existing movies without a version to standard version
+        for entry in self.cursor.fetchall():
+            self.add_videoversion(entry[0], entry[1], "movie", 40400, 40400)
```
oddstr13, line 181: What is the work that is intended to be done here?

Right now, it only loops over entries that don't have any entry in the videoversion, and as such it won't change 0 -> 1.

If I'm understanding it correctly, this code should only be needed when upgrading directly from older versions than omega?

oddstr13: This is what `videoversiontype` looks like on my Kodi 21.2 Windows install (videodb v131).

id | name | owner | itemType
-- | -- | -- | --
40400 | Standard Edition | 0 | 0
40401 | Extended Edition | 0 | 0
40402 | Unrated Version | 0 | 0
40403 | Uncut Version | 0 | 0
40404 | Remastered Version | 0 | 0
40406 | Theatrical Cut | 0 | 0
40407 | Director's Cut | 0 | 0
40408 | Special Edition | 0 | 0
40409 | Limited Edition | 0 | 0
40410 | Complete Edition | 0 | 0
40411 | The Final Cut | 0 | 0
40412 | Super Duper Cut | 0 | 0
40413 | Collector's Edition | 0 | 0
40414 | Ultimate Collector's Edition | 0 | 0
40415 | Criterion Collection Edition | 0 | 0
40416 | Fan Edit | 0 | 0
40417 | Black and White Edition | 0 | 0
40431 | 10th Anniversary Edition | 0 | 0
40432 | 20th Anniversary Edition | 0 | 0
40433 | 25th Anniversary Edition | 0 | 0
40434 | 30th Anniversary Edition | 0 | 0
40435 | 40th Anniversary Edition | 0 | 0
40436 | 50th Anniversary Edition | 0 | 0
40437 |   | 0 | 0
40438 |   | 0 | 0
... | ...  | 0 | 0

I think it would be best to read out the value and store in a Python variable, not sure of how big of a performance impact the additional select in the query has.

I also think you should just update the omega migration to also handle piers, they do the same thing, except the number has changed, but in your piers code, you now grab that number from the database, and that is also a valid strategy for the omega migration.

jellyfin_kodi/objects/kodi/queries.py
```diff
@@ -409,12 +409,12 @@
 add_set_obj = ["{Title}", "{Overview}"]
 add_video_version = """
 INSERT INTO     videoversion(idFile, idMedia, media_type, itemType, idType)
-VALUES          (?, ?, ?, ?, ?)
+VALUES          (?, ?, ?, (SELECT itemType FROM videoversiontype WHERE id = ?), ?)
```
oddstr13, line 412: What is the performance impact of this additional inner SELECT?

@coderabbitai maybe you can help with this question?
These are sqlite database queries.

jellyfin_kodi/objects/kodi/movies.py
pdf: >May or may not actually need to do this migration, as looking at the Kodi source, the migration is supposed to get applied on upgrade.

I suspect the reason for including the migration in the addon is to repair libraries that have been broken by the addon. The only other way for users to recover from having the current version of this addon running on Piers would be to delete their movie libraries entirely and sync from scratch, but users won't know to do that and so they will just have missing library items, since the upstream Kodi migration has already been run prio to the addon syncing broken items.

If this addon was updated before anyone ran it on Piers, then this wouldn't be an issue, but that's not where we are now.

JeroenED:
I will look into the changes this week. There is lot going on currently so I couldn't look into it yet

JeroenED:
Changes done and commited but still need to be tested.

oddstr13 (commented):
@coderabbitai review

oddstr13 (approved):

PR #1020, opened 2025-08-21
Test matrix additions

oddstr13:


oddstr13:
the reason for the requirements split is that python 3.8 doesn't have the pre-commit version requested

PR #1026, opened 2025-09-01
Modernise settings to format introduced with Leia

bossanova808:
**Changes**

* Modernise the settings file to the current xml format (this format was introduced way back with Kodi Leia
(New format: https://kodi.wiki/view/Add-on_settings_conversion - deprecated format: https://kodi.wiki/view/Add-on_settings)
* Silence (many, many!) log warnings by adding the missing, used settings: `LastIncrementalSync` `SyncInstallRunDone` `groupedSets` `platformDetected`
* Add resulting missing strings to `resources/resource.language.en_gb/strings.po` 
(new settings format requires _all_ labels to be in `strings.po`)
* (`git rm` the incorrectly cased `resource.language.en_GB` in favour of `resource.language.en_gb`) - it seems there are a few issues with capitals in the language section but have as yet only tackled this one....

**Testing**

I cross checked the settings (and resulting actual saved data in `addon_data`) and all seems to be ok.  I was able to set up and configure, in a variety of ways, several profiles.

I can't claim to have tested every setting/config pathway, of course.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -1020,3 +1020,156 @@ msgstr "Remote Path"
 msgctxt "#33207"
 msgid "Local Path"
 msgstr "Local Path"
+
+msgctxt "#33208"
+msgid "360"
+msgstr "360"
+
+msgctxt "#33209"
+msgid "480"
+msgstr "480"
+
+msgctxt "#33210"
+msgid "600"
+msgstr "600"
+
+msgctxt "#33211"
+msgid "720"
+msgstr "720"
+
+msgctxt "#33212"
+msgid "1080"
+msgstr "1080"
+
+msgctxt "#33213"
+msgid "Unlimited [default]"
+msgstr "Unlimited [default]"
+
+msgctxt "#33214"
+msgid "0.5 Mbps"
+msgstr "0.5 Mbps"
+
+msgctxt "#33215"
+msgid "1.0 Mbps"
+msgstr "1.0 Mbps"
+
+msgctxt "#33216"
+msgid "1.5 Mbps"
+msgstr "1.5 Mbps"
+
+msgctxt "#33217"
+msgid "2.0 Mbps"
+msgstr "2.0 Mbps"
+
+msgctxt "#33218"
+msgid "2.5 Mbps"
+msgstr "2.5 Mbps"
+
+msgctxt "#33219"
+msgid "3.0 Mbps"
+msgstr "3.0 Mbps"
+
+msgctxt "#33220"
+msgid "4.0 Mbps"
+msgstr "4.0 Mbps"
+
+msgctxt "#33221"
+msgid "5.0 Mbps"
+msgstr "5.0 Mbps"
+
+msgctxt "#33222"
+msgid "6.0 Mbps"
+msgstr "6.0 Mbps"
+
+msgctxt "#33223"
+msgid "7.0 Mbps"
+msgstr "7.0 Mbps"
+
+msgctxt "#33224"
+msgid "8.0 Mbps"
+msgstr "8.0 Mbps"
+
+msgctxt "#33225"
+msgid "9.0 Mbps"
+msgstr "9.0 Mbps"
+
+msgctxt "#33226"
+msgid "10.0 Mbps"
+msgstr "10.0 Mbps"
+
+msgctxt "#33227"
+msgid "12.0 Mbps"
+msgstr "12.0 Mbps"
+
+msgctxt "#33228"
+msgid "14.0 Mbps"
+msgstr "14.0 Mbps"
+
+msgctxt "#33229"
+msgid "16.0 Mbps"
+msgstr "16.0 Mbps"
+
+msgctxt "#33230"
+msgid "18.0 Mbps"
+msgstr "18.0 Mbps"
+
+msgctxt "#33231"
+msgid "20.0 Mbps"
+msgstr "20.0 Mbps"
+
+msgctxt "#33232"
+msgid "25.0 Mbps"
+msgstr "25.0 Mbps"
+
+msgctxt "#33233"
+msgid "30.0 Mbps"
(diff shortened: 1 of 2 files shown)


oddstr13:
I'm not entirely sure of *where* those incremental sync variables are supposed to be stored, I think they where possibly duplicated between our internal json config and the kodi-managed xml config last time I looked at some sync issues, but I haven't studied that in detail.

Just a note for later, not a change request for this PR in particular.

oddstr13 (dismissed):
The translation file changes are unrelated, undo those.

Translation is done using Weblate, and the second file seems to have been generated by it, so we need to figure out why and how to prevent it in the future.
Best just create an issue about it for now, and include details on what file name format is expected by Kodi if any.

The only language file you may ever touch in pull-requests is `resources/language/resource.language.en_gb/strings.po`, which is the source file for translation.

---

I have not yet tested the settings change. Does the actual setting format stay the same? Do we need to migrate settings, or does Kodi do that for us?

Before and after screenshots of the setting dialogs would be appreciated.

bossanova808:
The settings are identical after this change.  Kodi will re-write the user settings (with no changes to values).  

Behaviour before and after, both visually and functionally, is identical (as best as I can tell, again pretty much impossible to test _everything_ but given the actual user settings file is effectively identical, this should be very safe.

RE:

> The translation file changes are unrelated, undo those.

The strings have to be in the `.po` file, and the one I have edited is `resources/language/resource.language.en_gb/strings.po` so I am not sure what I can undo here?

Do you mean undoing the `git rm` of the incorrectly cased file?  If I don't do that, it shows up as a separate file, one that should not be committed, to anyone pulling the repo as it stands now.  I guess I could restore the file, and simply not have it as part of this commit, I suppose but it's in the git history and shouldn't be (it's a quite old version of the strings file as far as I can see, that somebody incorrectly committed with that casing long ago in the past).

If you look at this [history](https://github.com/jellyfin/jellyfin-kodi/blob/master/resources/language/resource.language.en_GB/strings.po) of this particular file, it was erroneously created and committed April 25 with this commit:
https://github.com/jellyfin/jellyfin-kodi/commit/7bebe3f9c2c6ebd05d33af555d7d1f998c5fe3eb

As far as I can tell, it really should be removed as I have done, as it just causes confusion.

The actual file is being edited in parallel here:
https://github.com/jellyfin/jellyfin-kodi/commits/master/resources/language/resource.language.en_gb/strings.po

oddstr13:
> Do you mean undoing the `git rm` of the incorrectly cased file? If I don't do that, it shows up as a separate file, one that should not be committed, to anyone pulling the repo as it stands now. I guess I could restore the file, and simply not have it as part of this commit, I suppose but it's in the git history and shouldn't be (it's a quite old version of the strings file as far as I can see, that somebody incorrectly committed with that casing long ago in the past).

Yes, please restore the file.

> If you look at this [history](https://github.com/jellyfin/jellyfin-kodi/blob/master/resources/language/resource.language.en_GB/strings.po) of this particular file, it was erroneously created and committed April 25 with this commit: [7bebe3f](https://github.com/jellyfin/jellyfin-kodi/commit/7bebe3f9c2c6ebd05d33af555d7d1f998c5fe3eb)
> 
> As far as I can tell, it really should be removed as I have done, as it just causes confusion.
> 
> The actual file is being edited in parallel here: https://github.com/jellyfin/jellyfin-kodi/commits/master/resources/language/resource.language.en_gb/strings.po

The existence of this file is just a symptom of a configuration issue somewhere.
It has been added via our Weblate instance, and could be re-created via it, so we should fix the underlying issue, and that is separate of the work otherwise being done here.

https://translate.jellyfin.org/browse/jellyfin/jellyfin-kodi/en_GB/?q=state:%3E=translated

If you look at the history for the file, you can see that it was created in April via Weblate, and was last updated a week ago.

https://github.com/jellyfin/jellyfin-kodi/commits/master/resources/language/resource.language.en_GB/strings.po

Additionally, Weblate tends to throw a fit every time translations other than the base translation are modified directly.

bossanova808:
Ok, have removed the `git rm` from this PR.  

This PR is now just the settings format changes (which result in no visible or behaviour changes) - and the missing string being added in the one file mentioned above (i.e. `resources/language/resource.language.en_gb/strings.po`)



**Not related directly to this PR, therefore, but for further info:**

That was quite challenging, basically impossible to do on Windows (where I normally do my Kodi dev) - due to the lack of a cased file system...so moved things over to WSL to tidy it up.  On Windows, only the `en_GB` file gets pulled, and it therefore immediately shows up as a commit-able change on a fresh clone - behaviour I have never seen before - you get this warning:

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'resources/language/resource.language.en_GB/strings.po'
  'resources/language/resource.language.en_gb/strings.po'
```

The only case _overlap_ I can see it this one:

image

..but for consistency, I'd suggest also fixing up the casing of these two as otherwise there's bound to be similar issues arising with them at some point:

image

image

(I would ague the repo should be made multi-platform friendly by tidying this up ASAP, as that will help encourage contributions without people running in to these confusing issues that require a reasonable amount of platform and git fu to solve).

oddstr13:
@coderabbitai fullreview
Compare old and new setting formats and look for inconsistencies / likely unintentional changes.

oddstr13:
@coderabbitai in the `settings.xml`, `<dependency>` of `type` `visible` or `enable` with a `infobool` `condition` with a value of (`true`, `false`, `yes` or `no`) should use tags `<visible>` or `<enable>` with a value of `true` or `false` directly instead.

Further, both of them default to `true`, so it would likely be best to omit it completely when this is the case (only needed when the setting should always be disabled or hidden).

bossanova808:
Thanks for all your extra work here and apologies that extra work was needed!

(I'm new to this AI coding thing...yesterday we had no power here (works in our street) so I woke up this morning to like 50 rabbit messages...amazing where this has all got to!!)

PR #1044, opened 2025-09-15
FEATURE: Making _convert_endpoint_address_to_manual_address IPv6 friendly

donneypr:
Hello,

I was having a look at jellyfin_kodi/jellyfin/connection_manager.py and it seems like _convert_endpoint_address_to_manual_address is only accepting IPv4 addresses. There was a TO-DO comment to make it IPv6 friendly so I've added to the function to fix that. 

Please let me know what you think. 

## Summary by CodeRabbit

- New Features
  - Added IPv6-aware server address handling, including bracketed and non-bracketed formats, with canonical formatting.
- Bug Fixes
  - Corrected host/port parsing across IPv4 and IPv6 formats to prevent misdetection and connection issues.
  - Improved validation to reject incomplete or invalid addresses instead of proceeding with faulty connections.
  - Reduced errors by consistently normalizing addresses before use.

### DIFF
--- a/jellyfin_kodi/jellyfin/connection_manager.py
+++ b/jellyfin_kodi/jellyfin/connection_manager.py
@@ -270,24 +270,43 @@ class ConnectionManager(object):
 
         return servers
 
-    # TODO: Make IPv6 compatible
     def _convert_endpoint_address_to_manual_address(self, info):
+        addr = info.get("Address")
+        endpoint = info.get("EndpointAddress")
+        if not addr or not endpoint:
+            return None
+        # Getting the host from EndpointAddress checking for IPv6 formatting
+        if endpoint.startswith('['):
+            rb = endpoint.find(']')
+            if rb != -1:
+                host = endpoint[1:rb]  
+            else:
+                return None  
+        else:
+            if endpoint.count(':') > 1:
+                host = endpoint
+            else:
+                host = endpoint.rsplit(':', 1)[0] if ':' in endpoint else endpoint
 
-        if info.get("Address") and info.get("EndpointAddress"):
-            address = info["EndpointAddress"].split(":")[0]
+        port = None
+        if addr.startswith('['):
+            rb = addr.find(']')
+            if rb != -1 and rb + 1 < len(addr) and addr[rb + 1] == ':':
+                ps = addr[rb + 2:]
+                if ps.isdigit():
+                    port = int(ps)
+        else:
+            if addr.count(':') == 1:
+                ps = addr.rsplit(':', 1)[1]
+                if ps.isdigit():
+                    port = int(ps)
 
-            # Determine the port, if any
-            parts = info["Address"].split(":")
-            if len(parts) > 1:
-                port_string = parts[len(parts) - 1]
+        if port is None:
+            return None
 
-                try:
-                    address += ":%s" % int(port_string)
-                    return self._normalize_address(address)
-                except ValueError:
-                    pass
+        combined = f"[{host}]:{port}" if ':' in host else f"{host}:{port}"
+        return self._normalize_address(combined)
 
-        return None
 
     def _normalize_address(self, address):
         # TODO: Try HTTPS first, then HTTP if that fails.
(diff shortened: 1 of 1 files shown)


oddstr13:
This looks overly complex, and also changes behavior when urllib3 throws ValueError.

I also spotted it when looking thru TODOs while working on some cleanup the other day, and thought it'd make sense to look for some stdlib functionality to replace the naive split, as that was just a quick and dirty solution then and there.

@coderabbitai fullreview

Supproted Python version range: 3.8 to 3.13

donneypr:
Thanks for the feedback @oddstr13 , I can do some investigation on stdlib and see how we can implement it to make it IPv6 friendly. Will update here on any progress.

donneypr:
Will have a look at CodeRabbits feedback and see what we can do. I believe we can use urllib3 instead of urllib right @oddstr13 ?

oddstr13:
@coderabbitai that still looks very unoptimal, keep in mind that a lot of this is already done in _normalize_address

@donneypr @coderabbitai This code was written back when the code-base was also supporting Python 2.7, and that is likely a large part of why it was this hacky in the first place. This change requires careful review of how those two functions are getting used, and what the intended effect is. It has been a long time since I was working in those parts of the code-base.

urllib3 is already being used in the code-base, so that should be fine, but stdlib might also have the desired functionality.

It might make sense to merge those two functions, but it also might not make sense.

I believe we are likely working on a manually entered address at this point, which we are trying to magic into what the user actually meant to type. It might also already be a fully qualified address from the auto-discovery mechanism.
The two TODOs here hints at which future improvements where envisioned when the code was (last re-)written.

Scenario:
User enters "jellyfin-server", and we need to figure out whether that is http://jellyfin-server:8096, https://jellyfin-server/ or something else (Jellyfin default ports are 8096 for http and 8920 for https). We could also choose to just error, expecting the user to always enter a full url.


In the future, I would like to also implement (m)DNS-SD, but that's out of scope here.

It would make sense to implement a proper test-suite before changing these functions, as they don't need access to any Kodi internals. Would have to convert them to classmethod/staticmethod first tho.

oddstr13:
@coderabbitai update learnings to correct default https port to 8920

donneypr:
Thank you for the insight @oddstr13 . Would a good start to tackling this issue be:

1. Making the current _convert_endpoint_address_to_manual_address & _normalize_address classmethod/staticmethod first.
2. Create a robust test suite to work out the behaviour of these methods before we change anything.
3. Look into (m)DNS-SD and stdlib to make the connection_manager.py IPv6 friendly.

What do you think?

oddstr13:
DNS-SD is out of scope here, and would require mocking to test properly.

I need to ask; How much AI are you using for generating code and comments, and how much of this code do you, yourself understand?

oddstr13:
I'm not saying all AI is bad (as you can see, I am experimenting with AI reviews), but when you start generating code, you take on the full responsibility of a senior developer / team lead, supervising a fresh junior developer.

When submitting pull-requests it is important that you disclose the fact, if you are using AI and maybe not fully understanding what is going on.

If you yourself are not providing any additional input, but just asking the AI to generate a solution to an issue/task, and passing my feedback to AI and giving me the AI response, then you are not providing any value in the exchange.
It would be more efficient for me to talk to the AI directly, with lower latency between questions/replies – but I would generally also use the AI more for review/sanity checking, and the research phase, not for the actual implementation.

I'm not saying any of this to cast shame or to be mean, but to provide my views on the topic, so that you might learn from it and improve your programming skill.

Reviewing other people's code can be quite a bit of work depending on the change, and even more so when you have to guide new developers (new to the codebase, language or programming), but it is still something I do (not all the time, because it is hard work and can at times be frustrating), because we all have to start somewhere, and I was once the kid asking all the (at times annoyingly simple) questions.
This all tend to depend on the willingness to learn, however, as trying to teach someone who are not interested in learning is a futile endeavor.

I hope this experience doesn't discourage you from learning programming :slightly_smiling_face:

Issue #912, opened 2024-09-08
Resuming media plays from incorrect time


ChocOranger:
**Describe the bug**
When resuming a movie the time displayed in Kodi's context popup for resume is correct (eg Resume from 00:13:45) will actually resume from around a minute or so before that time (eg 00:12:45). Actual time appears to vary depending on how far played (eg  1:08:31 resumes @ 1:07:07)
The jumpback on resume time is the default of 10 seconds.
I originally posted on the Jellyfin forum https://forum.jellyfin.org/t-jellyfin-kodi-addon-resuming-media-jumps-back-about-a-minute with one other person replying they had the same issue. Nobody replied it was working correctly for them.

**To Reproduce**
1. Play a movie
2. Stop the movie part way through noting the time played
3. Restart the movie and the resume from should be the time played in step 2 minus jumpback time.
4. Actual restart position is around a minute earlier than displayed in the Kodi context menu.

**Expected behavior**
Movie restarts from the stopped time minus the jumpback on resume time.

**Logs**
Couldn't see any relevant log errors

**Screenshots**

**System (please complete the following information):**
All versions latest stable available. Clean installs of portable versions to test
- OS: Windows 10. Also happens on Shield Pro.
- Jellyfin Version: 10.9.9
- Kodi Version: 21.1 Portable. Default skin with no extra addons.
- Addon Version:  10.0.4 + py3
- Playback Mode: Add-On

**Additional context**
Kodi without jellyfin resumes correctly.
Jellyfin web interface resumes correctly.
Jellyfin addon causes a jumpback of around a minute from the time shown in Kodi's context window.

sic6SaNdMaN:
I can confirm this problem.
It occurs to me since I switched from "native paths" to "AddOn mode" due to this discussion:
https://github.com/jellyfin/jellyfin/discussions/11654

I never switched back to "native paths" because I want to test "AddOn mode" more, if I can find anymore issues.
And I think, that this IS an issue with AddOn mode - I never experienced this behavious with native paths.

sic6SaNdMaN:
More Info: I'm pretty sure, that the resuming time is always exactly one minute before the actual stopping time.
So the configured time isn't used, it's always 60 seconds instead.

bossanova808:
[...]

PR #1051, opened 2025-10-11
Remove setting of `StartPercent` 

bossanova808:
Setting `StartPercent` and with a random 0.4% offset makes no sense given an accurate `resumetime` is available.

See discussion here: https://github.com/jellyfin/jellyfin-kodi/issues/912#issuecomment-3392922726

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video resume behavior: playback now relies on the saved resume time without using or showing an inconsistent start percentage, delivering more consistent start positions and clearer resume indicators. When not resuming, videos start from the beginning as expected.

### DIFF
--- a/jellyfin_kodi/objects/actions.py
+++ b/jellyfin_kodi/objects/actions.py
@@ -508,9 +508,6 @@ class Actions(object):
 
             if obj["Resume"] and item.get("resumePlayback"):
                 listitem.setProperty("resumetime", str(obj["Resume"]))
-                listitem.setProperty(
-                    "StartPercent", str(((obj["Resume"] / obj["Runtime"]) * 100) - 0.40)
-                )
             else:
                 listitem.setProperty("resumetime", "0")
                 listitem.setProperty("StartPercent", "0")


oddstr13:
@coderabbitai fullreview

What purpose could setting both resumetime and StartPercent here have?
What are each of those used for internally in Kodi (and maybe other places in this addon)?
What effects could this change have?

bossanova808:
I suspect that _maybe_ this line points to the root cause:
```
        is_video = obj["MediaType"] in ("Video", "Audio")  # audiobook
```
I think perhaps `StartPercent` came along much later with AudioBook support (not sure why - but that's the only place I see any discussion about it - although I am out and about at the moment and annoyingly can't find that discussion again).

I think it's pretty clear, for video playback, that it's a significant bug, as it stands, anyway.  

(Moving away from `listitem.setWhatever` to `InfoTagger` is a significantly bigger thing - definitely do-able (and probably will _need_ to be done at some point) - but that would require a fair bit more understanding & then considerable testing to tackle properly).  

I think consistently inaccurate resumes for videos, in the context of Kodi & Jellyfin, is simply too big a price to pay for whatever this might improve for AudioBooks - which I suspect _very_ few Kodi using people actually use.    Of course one could gate things with an `if  obj["MediaType"] == "Audio" ... StartPercent else ... resumetime` approach.  That's probably likely to be the least risky approach...

oddstr13:
Audiobook support is rudimentary at best, and Jellyfin isn't the best manager for that in the first place, I don't think we should worry too much about that breaking.
I don't see a way to sync audiobooks in the first place, and the dynamic menu currently time out for me due to slow processing server side and the amount of items.

bossanova808:
@oddstr13 

Ok have been real world testing this the last few days and it's without question a positive change

```
### Jellyfin Fixer 0.0.4: Currently playing time: 1691.775, Kodi resume: 1693.0, difference: 1.2
### Jellyfin Fixer 0.0.4: Currently playing time: 1187.7830000000001, Kodi resume: 1189.0, difference: 1.2
### Jellyfin Fixer 0.0.4: Currently playing time: 1535.832, Kodi resume: 1538.0, difference: 2.2
### Jellyfin Fixer 0.0.4: Currently playing time: 1373.837, Kodi resume: 1374.0, difference: 0.2
### Jellyfin Fixer 0.0.4: Currently playing time: 1423.775, Kodi resume: 1424.0, difference: 0.2
### Jellyfin Fixer 0.0.4: Currently playing time: 1942.83, Kodi resume: 1945.0, difference: 2.2
### Jellyfin Fixer 0.0.4: Currently playing time: 793.936, Kodi resume: 796.0, difference: 2.1
```

All of those resumes are within normal expectations (this is with JF for Kodi set to a 0 resume offset, and the minor differences here are presumably accounted for by when exactly `onAVStarted` is called after playback begins...)

Since this change, my fixer addon has not stepped in once to fix things (as by default it allows a time difference delta of 7 seconds between expected and actual)...which is ideal, of course, and demonstrates this fix does what it should.  

I suspect I will still see the occasional reset-to-zero I have occasionally seen (and my fixer addon would still intercept and fix) - but that appears to be a different issue.   

And I agree that we shouldn't worry about Audiobooks (nor am I convinced this would actually cause any harm there either as again, the whole thing is guarded by `obj["Resume"]` being available.

oddstr13 (approved):

PR #1060, opened 2025-10-25
Add 0001-prefer-local-server-address.patch by upload

trumblejoe:
Prefer local server address patch allows the server to use both a local server address or remote server address through server discovery or manual input from the user. Local address will be preferred then remote address.

### DIFF
--- a/0001-prefer-local-server-address.patch.txt
+++ b/0001-prefer-local-server-address.patch.txt
@@ -1,0 +1,1090 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: trumblejoe <trumblejoe@example.com>
+Date: Fri, 25 Oct 2025 17:50:01 +0000
+Subject: [PATCH] Prefer local server address and allow manual override
+
+Add automatic detection of local vs remote server address, quick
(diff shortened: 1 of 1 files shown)


mcarlton00:
I disagree with including this feature at all, and we've explicitly removed similar functionality from the client years ago because it was more trouble than it was worth and caused extra support burden due to user error.  This is the entire point of local DNS, which is a well defined standard and works automatically instead of adding code complexity for a niche feature.

Perhaps more importantly, this is an invalid PR.  There is absolutely no reason for you to ever provide a patch file on github like this.  It's a a waste of everyone's time and makes reviewing it 1000% part harder than it needs to be.  If you want to change a file, change the file.  Don't add an entirely new file telling me how to change other files.

trumblejoe:
Local dns isn't supported on most consumer routers and relying on a self hosted DNS service is just another external dependency.

If you had issues with the feature you should have commented on the issue raised over three weeks ago. Githubs PR feature doesn't allow uploading a standalone patch file, forking it and issuing it through this is the only way to get a patch file sent over, short of adding it to the issue which I did and which no one looks at.

Issue #1073, opened 2025-11-26
"Next Episodes" no longer works (empty)


hjpaul7:
### Summary

The "Next Episodes" no longer works in the addon. It just shows an empty directory.

Jellyfin "Next Up" (via web UI) shows next up episodes just fine.

Server: 10.11.3

### Steps to Reproduce

1. Launch Jellyfin addon
2. Navigate to Shows -> Next Episodes
3. Observe empty directory 

### Operating System

Arch Linux

### Jellyfin Server Version

10.11.3

### Addon Mode

Addon-mode

### Logs

No logs

joopdo:
Hi @hjpaul7 , maybe related.

I have a problem in Kodi, where one Jellyfin library doesn't show any items in it (documentaries) - but it does work in the app. I tried a lot of things. Finally, I made a new Jellyfin account and reset Jellyfin in Kodi, and logged into the new account. To my joy, the missing library did show items now! However, in this account, 'next episodes' doesn't seem to work. Initially, I disabled the view, thinking it will populate, but 3 weeks later, still nada. 

So maybe it's because you have made an account in a recent version, which has this broken? Just thinking out loud here...

hjpaul7:
> Hi [@hjpaul7](https://github.com/hjpaul7) , maybe related.
> 
> I have a problem in Kodi, where one Jellyfin library doesn't show any items in it (documentaries) - but it does work in the app. I tried a lot of things. Finally, I made a new Jellyfin account and reset Jellyfin in Kodi, and logged into the new account. To my joy, the missing library did show items now! However, in this account, 'next episodes' doesn't seem to work. Initially, I disabled the view, thinking it will populate, but 3 weeks later, still nada.
> 
> So maybe it's because you have made an account in a recent version, which has this broken? Just thinking out loud here...

Interesting thought, however it's the same account for about a year now. So you're still having issues with "Next Episodes" in the new account?

I wondering if Jellyfin made changes to "Shows" library type in 10.11 that may be conflicting here in the Kodi addon.

joopdo:
That is correct, "Next Episodes" does not work with the new second account. If, however, I login to the older account again on Kodi, "Next Episodes" does work, but then I miss my Documentaries Library.

joopdo:
I tried again and indeed 'next episodes' is empty for both accounts. For now I solved it by migrating to JellyCon.

hjpaul7:
[...]

PR #1075, opened 2025-11-29
Fix: Next Episodes (maxDaysInNextEpisodes)

hjpaul7:
Fixes https://github.com/jellyfin/jellyfin-kodi/issues/1073

More reports: [Reddit thread](https://www.reddit.com/r/jellyfin/comments/1oob1fr/kodi_plug_in_next_up_not_working_across_all_of_my/)

- Missing `maxDaysInNextEpisodes` in settings.xml.
- Confirmed working with my Kodi setup


h3llrais3r:
It seems that the setting was lost with https://github.com/jellyfin/jellyfin-kodi/pull/1026
Tested on my laptop installation and this PR works.
Thanks for bringing it back!

hjpaul7:
Anyone?

@oddstr13 You seem to be the active maintainer

phamaralbr:
Applied your fix and was able to get some entries on "Next Episodes", but they don't really make sense. Is something still wrong or am i just misunderstanding what should show up here?

To the left, I have the jellyfin dashboard showing the next episodes to the shows i'm watching. 
But to the right, jellyfin for kodi shows results that don't really make sense, stuff I've already watched. Shouldn't I get the same entries i get on Jellyfin?

image

solazs:
@phamaralbr I guess that would depend on what you set the setting to.

Issue #549, opened 2021-07-14
Cinema mode not working only intro plays


mLgz0rn:
**Describe the bug**

So I've tried to use the cinema mode, and it only seems to play the intro, it never continues to play the movie/show

**To Reproduce**

1. Install intros plugin on Jellyfin
2. Enable Cinema mode in Jellyfin settings on Kodi
3. Play movie or show
4. Intro now plays
5. When intro is done nothing else is playing and it just returns to the menu

**Expected behavior**

Play the intro then continue to play the movie/show after

**Logs**

[Log](https://paste.kodi.tv/etubirolug)

**System (please complete the following information):**
 - OS: Android TV
 - Jellyfin Version: 1.0.7.6
 - Kodi Version: 19.1
 - Addon Version: 0.7.4+py3
 - Playback Mode: Add-On

Extarys:
Side note: I think I have the same issue with android Jellyfin and external player but I need to do more testing. I definitly have the issue you describe for kodi though. I'm not sure how it works but I believe each individual client need to have support for the intro plugin for it to work properly.

oddstr13:
I can confirm that this is indeed an issue, no idea of what changed to cause it, however.
Workaround is to disable cinema mode in the addon.

mcarlton00:
As an update: we currently have no method of playing more than one file, it just plays whatever is first up in the queue. https://github.com/jellyfin/jellyfin-kodi/blob/d69f96f9f3a9aacaf5a95d83f8857eae23c4676e/jellyfin_kodi/objects/actions.py#L78-L79

I tried copying the playback commands used in JellyCon because intros work fine there (adding things to a playlist and then `xbmc.Player().play(playlist)`), and it instead caused Kodi to completely lock up and eventually crash back to the desktop.  So it looks like we're going to have to rework the playback code somehow.  In the meantime, we should probably change the settings file to have cinema mode off by default on new installs.

tehguitarist:
I’m a novice but I’ll see if I can find a function from this kodi addon: https://github.com/CinemaVision/script.cinemavision
which also handles pre rolls that might help.

tehguitarist:
Ok so i've done some digging, it looks like the fuction they use is on line 887 of https://github.com/CinemaVision/script.cinemavision/blob/master/lib/experience.py

[...]

PR #1077, opened 2025-12-06
Fix: Cinema mode with playlist method

hjpaul7:
Fixes: https://github.com/jellyfin/jellyfin-kodi/issues/549

Tested on local machine and Ugoos, haven't had any issues so far but other testers would be great.


Beormund:
I have been using the fix for several months - it works well.

mcarlton00 (approved):

Issue #1084, opened 2025-12-15
Issue: strm file with local file path fails to play


VergilGao:
https://github.com/jellyfin/jellyfin-kodi/blob/cee3ef3a95235c584cd606b560154322a0d8c25f/jellyfin_kodi/helper/playutils.py#L198

**Problem Description**
When a strm file contains a local file path (instead of an HTTP URL), the Jellyfin for Kodi plugin attempts to play the local file directly, resulting in playback failure.

**Steps to Reproduce**
Create a strm file with content pointing to a local path:
`/mnt/rclone/media/public/shows/Stranger Things/Season 01/S01E01.mp4`
Attempt to play this strm file through the Jellyfin plugin in Kodi

Playback fails with errors indicating the plugin is trying to access the local file path directly

**Expected Behavior**
strm files should play correctly regardless of whether they contain HTTP URLs or local file paths.

**Actual Behavior**
For strm files containing local paths, the plugin attempts to play the file directly via the local filesystem, causing errors:
```
2025-12-15 19:32:13.781 T:7288     info <general>: VideoPlayer::OpenFile: plugin://plugin.video.jellyfin/fc8fa12ee6bc<token ...>/52b0b62b<token ...>/?filename=Stranger+Things+S01E07.strm&id=fd71458a5d03<token ...>&dbid=50&mode=play
2025-12-15 19:32:13.783 T:9924     info <general>: Creating InputStream
2025-12-15 19:32:13.783 T:9924    error <general>: GetDirectory - Error getting /mnt/rclone/media/public/shows/Stranger Things/Season 01/
2025-12-15 19:32:13.798 T:9924    error <general>: CVideoPlayer::OpenInputStream - error opening [plugin://plugin.video.jellyfin/fc8fa12ee6bc<token ...>/52b0b62b<token ...>/?filename=Stranger+Things+S01E07.strm&id=fd71458a5d03<token ...>&dbid=50&mode=play]
2025-12-15 19:32:13.798 T:9924     info <general>: CVideoPlayer::OnExit()
```

oddstr13:
I'm not intimately familiar with the particulars of strm files, but this kinda sounds like expected behavior to me?

You might want filesystem symbolic links in stead?

VergilGao:
> I'm not intimately familiar with the particulars of strm files, but this kinda sounds like expected behavior to me?
> 
> You might want filesystem symbolic links in stead?

[...]

PR #1085, opened 2025-12-16
fix: strm file with local file path fails to play

VergilGao:
this should fix https://github.com/jellyfin/jellyfin-kodi/issues/1084

I will continue testing for a few more days. If no issues arise, I will update this PR.


VergilGao:
it works fine on my `coreelec 21.3 ng` system.

mcarlton00 (approved):
I have no good way of testing this right now, but the code seems to make sense

Issue #953, opened 2024-12-16
Implement Jellyfin's new segment API (intro skipping)


crylcy:
I've noticed that while watching TV series on the Jellyfin web interface, there's an option to skip intros and outros, but this feature is missing in Kodi. I hope that a future update to the Kodi plugin could include the ability to skip intros and outros, similar to the functionality available in Jellyfin web interface。

Thanks a lot~！

dgbdgb264:
+1 for this request

adriano-poci:
With the deprecation of the intro-skipper plugins (which used to work with Jellyfin-Kodi), and Jellyfin's new native implementation which is client-side, the only way to get that functionality back is to implement it directly in the Jellyfin-Kodi plugin as a playback option.

+1 for this request.

oddstr13:
Can you link references for the intro skipper deprecation? issue/blog/forum-post or such, explaining the why & when would be nice

Also, please use the thumbs up reaction on the initial issue rather than comments with +1/me too, github allows sorting by number of reactions :)

theinfguy:
+1 from me. It would be great to see Jellyfin for Kodi implement an intro/outro skip control based on the new "Media Segment" API thats in JF 10.10:

https://features.jellyfin.org/posts/45/skip-intro-outro-credits-feature

SgtJalau:
A very cool feature, that I would love to see included. Well, I love it that much that I created a little standalone addon. It's far from fancy and my first "proper" addon. It also probably still has quite a few bugs and issues, but it worked decently enough:

https://github.com/SgtJalau/service.jellyskip

TheOverpassArsonist:
> With the deprecation of the intro-skipper plugins (which used to work with Jellyfin-Kodi), and Jellyfin's new native implementation which is client-side, the only way to get that functionality back is to implement it directly in the Jellyfin-Kodi plugin as a playback option.
> 

fwiw there's (I think?) a fork of Intro-Skipper that seems to be getting maintained. (https://github.com/intro-skipper/intro-skipper) I haven't managed to find the time to get it fully setup yet, but I can't see any reason to believe it is not being maintained. 

With that said, does Kodi itself have anything builtin that would enable this feature to be added reasonably? I can't recall ever seeing a popup over the video player that had actual functionality. (of course, I suppose without using something like this to convey the segment data in the first place I guess there isn't any reason I _would_ have seen it either) Is this a glue-job or would implementing this require designing a custom interface element?

[...]

PR #1089, opened 2026-01-01
feature: mediasegments skip

ticoombs:
# MediaSegments (Skip?)

This provides a "Skip {segment} (time)" box, with a "Close" option next to it in the bottom right hand corner. 
It attempts to be smart and when you are _in the segment_, it will offer to skip. This allows for seeking into a segment as opposed to watching a segment. 

As you can see from the commits it was originally for the intro-skipper plugin (which I am using to generate the 'media-segments') but only realised after that there is no need to 'fallback' to the mediasegments, but just query it first. 

I took inspiration from the `up-next` addon with how they managed to create their popup. Thus it might be highly likely it looks very similar. 

Tested on a rPi 4 running LibreElec.

Items that i think need reviews on: 
- The UI around the XML / Skip Button.
  - I couldn't get the bounding to be variable? So if/when there are multiple languages the box isn't going to increase. And i'm not sure how to handle it. 
- Removal of the (possibly overly verbose) debug logging? (It certainly was very very useful in figuring it out, but might be too verbose)
- I made it for me, but figured it might be useful for others. And moved it to Media Segments so it should be supported by any app which updates Jellyfin's mediasegment.

(Disclaimer: I got some help with some AI doing it, but modified it myself too, if this is against a CoC, feel free to close :) )

fixes: #953

### DIFF
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -355,6 +355,9 @@ class Player(xbmc.Player):
         if window("jellyfin.external.bool"):
             return
 
+        if settings("mediaSegmentsEnabled.bool"):
+            self.check_skip_segments(item, item["CurrentPosition"])
+
         if not report:
             previous = item["CurrentPosition"]
 
@@ -381,9 +384,6 @@ class Player(xbmc.Player):
 
                 self.up_next = True
                 self.next_up()
-
-            if settings("mediaSegmentsEnabled.bool"):
-                self.check_skip_segments(item, item["CurrentPosition"])
 
             if (item["CurrentPosition"] - previous) < 30:
                 return
@@ -412,9 +412,6 @@ class Player(xbmc.Player):
             "SubtitleStreamIndex": item["SubtitleStreamIndex"],
         }
         item["Server"].jellyfin.session_progress(data)
-
-        if settings("mediaSegmentsEnabled.bool"):
-            self.check_skip_segments(item, item["CurrentPosition"])
 
     def onPlayBackStopped(self):
         """Will be called when user stops playing a file."""


jumoog:
nice work

sneakykiwi:
Hopefully this gets merged soon

jumoog:
Is there something missing or do you need help?

mihawk90:
Since you didn't mention it, is there a (configurable) auto-skip option akin to the webUI settings? 🤔

ticoombs:
> is there a (configurable) auto-skip option

There is now. It follows the jellyfin pattern. `Ask, AutoSkip, Nothing` now. And each Type has that option too. 

For those who can't wait for it to be merged/test it on your own kodi/libreelec boxes, rsync it to `/storage/.kodi/addons/plugin.video.jellyfin/`, restart kodi, and then check your Jellyfin Addon settings.

jellyfin_kodi/dialogs/skip.py
```diff
@@ -0,0 +1,118 @@
...
+from ..helper import LazyLogger
+from ..helper.translate import translate
+
+LOG = LazyLogger(__name__)
+
+# Action IDs
+ACTION_PARENT_DIR = 9
+ACTION_PREVIOUS_MENU = 10
+ACTION_BACK = 92
+ACTION_SELECT = 7
```
mcarlton00: This variable is never used

jellyfin_kodi/dialogs/skip.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
+from __future__ import division, absolute_import, print_function, unicode_literals
+import xbmcgui
+from ..helper import LazyLogger
+from ..helper.translate import translate
```
mcarlton00: This import doesn't look like it's being used anywhere.

jellyfin_kodi/player.py
```diff
@@ -20,11 +20,16 @@
 
 #################################################################################################
 
+SETTING_MEDIA_SEGMENTS_ENABLED = "mediaSegmentsEnabled.bool"
```
mcarlton00: I'm not sure I understand the benefit of setting a global variable for this string.  Especially since the variable name is longer than the string it's representing, it's not like you're saving keystrokes or anything.  Plus afaik in the rest of the codebase we just call the setting name directly instead of abstracting it away like this.

jellyfin_kodi/player.py
```diff
@@ -472,3 +507,194 @@ def stop_playback(self):
...
+            "Commercial": "skipCommercialMode",
+        }
+        setting_key = setting_map.get(segment_type)
+        if not setting_key:
+            return 0
+        return int(settings(setting_key) or 0)
+
+    def _handle_skip_segment(self, segment_type, start, end, mode):
+        LOG.debug("_handle_skip_segment: type=%s, mode=%d, start=%.1f, end=%.1f",
+                 segment_type, mode, start, end)
```
mcarlton00: This line is underindented according to the linter

jellyfin_kodi/dialogs/skip.py
```diff
@@ -0,0 +1,118 @@
...
+    "Recap": "Recap",
+    "Preview": "Preview",
+    "Commercial": "Ad",
+}
+
+
+class SkipDialog(xbmcgui.WindowXMLDialog):
+    """
+    OSD overlay dialog for skipping intro/outro segments.
+
```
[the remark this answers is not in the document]
ticoombs: Too used to my pre-commit's sorting it for me :sweat:

ticoombs:
Thanks for helping with it!

trollollol123:
Shouldn't it be checked more often for intro? Why not check every 0.5s for the first 10 min for an intro? That way you don't have to end up watching like 10s sometimes of the intro before it realizes it should skip. I'm using autoskip.

Also, there should be an option to delay the skip for x seconds. I like to see a second or two of the intro before it skips; that way I can feel safe that it's actually skipping the intro and not something else by mistake.

mihawk90:
Commenting on merged and closed PRs is pointless since the change is already made.

#1104

PR #1095, opened 2026-02-05
Add "Jellyfin transcode" context menu item to Live TV section

### LINKED ISSUE #1088: Transcode Option Missing From Live TV Context Menu
### Summary

Live TV only works for me when transcoded. There's no context menu for transcoding in the Live TV section so Force Transcode has to be chosen in the addon settings for all content.

I don't know why transcoding is necessary as Kodi can direct play these streams but while it's necessary it'd be good to have the context menu option for transcoding so other content can be left to direct play.

### Steps to Reproduce

Enable the option to transcode in the addon settings and navigate to a Live TV channel. There is no transcode option in the context menu.

kontell:
Tested and working on V21 & V22.

Fixes: https://github.com/jellyfin/jellyfin-kodi/issues/1088

### DIFF
--- a/.build/template.xml
+++ b/.build/template.xml
@@ -23,6 +23,7 @@
 				<label>30402</label>
 				<visible>[[!String.IsEmpty(ListItem.DBID) + !String.IsEqual(ListItem.DBID,-1) |
 					!String.IsEmpty(ListItem.Property(jellyfinid))] + [String.IsEqual(ListItem.DBTYPE,movie) |
+					String.Contains(Container.FolderPath,livetv) |
 					String.IsEqual(ListItem.DBTYPE,episode)]] +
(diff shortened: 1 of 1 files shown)


mcarlton00:
This fixes the missing option in the context menu, but playback seems to be failing when using it.  Do you have any interest in looking into that as well or only the context menu?  (either answer is fine, I'll sort out the playback issues if you don't want to)

kontell:
Playback's working for me without doing anything else. Is it working for you with force transcode set from the addon settings?

mcarlton00:
Also no.  I found the problem, seems to be related to the various types of live tv sources and how they get interpreted.  I'll PR my fix separately since it's unrelated.

mcarlton00 (approved):

PR #1097, opened 2026-02-10
Migrate to current auth methods

### LINKED ISSUE #1091: api_key removed from Jellyfin API, use ApiKey instead
See jellyfin/jellyfin#13306 and jellyfin/jellyfin#15559.

This will cause a hard-break in backwards compatability for versions prior to 10.8, which is when ApiKey was added.

Also review all other authentication related code, there are more changes.

mcarlton00:
Niels I heard you the first 300 times, it's quite rude to keep reminding me.

Fixes #1091 
* `api_key` to `ApiKey`
* old multi-header format to Authorization

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -305,7 +305,7 @@ class PlayUtils(object):
         self.info["Method"] = "DirectStream"
 
         if self.item["Type"] == "Audio":
-            self.info["Path"] = "%s/Audio/%s/stream.%s?static=true&api_key=%s" % (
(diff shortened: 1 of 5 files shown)


oddstr13:
need to pull or update some of those language linting tests, weblate keeps breaking them.
only actual CI failure here is missing black autoformat as far as I can tell

oddstr13 (approved):
Works fine in my limited testing

oddstr13:
For sure breaks backwards compat with Jellyfin 10.7 and older, 10.8 and up should be fine.

PR #1106, opened 2026-03-14
Map movie extras into Kodi database on full refresh and partial sync.

### LINKED ISSUE #787: Feature Request: Expose Extras
Jellyfin already has the ability to scan and sort [various types of extras](https://jellyfin.org/docs/general/server/media/movies#movie-extras) into the library given they are in the correct directories, and they are appropriately displayed in the WebUI.
In Kodi it seems there is no way to expose this functionality though.

After installing the [Extras Addon](https://kodi.wiki/view/Add-on:Extras) (and restarting Kodi), the Extras option is available, however it has no content (movie plays fine though, so share is reachable). Switching to Native Playback mode unfortunately had no effect at all.

Would be nice to see this at one point :)

midfords:
Resolves #787

This maps all movie extras from Jellyfin into the Kodi database. It maps in on full library refresh and with regular sync updates. Extras are accessible through the movie "information" page "Extras" button. 

Note: This is not compatible with the "Extras" optional add on for Kodi, as this plugin only works with local media files. It is only compatible with the new integrated "Extras" button on the movie information page.


dan-riley:
I just finished a version of this as well (actually finished it last week but just got around to cleaning up, but not ready for a pull request yet).  Mine does Video Versions primarily, as well as extras.  https://github.com/dan-riley/jellyfin-kodi/releases

PR #1125, opened 2026-04-04
Fix: Sync theme media for TV Tunes add-on

### LINKED ISSUE #1113: Sync theme media can't find TV Tunes
### Summary

I didn't even know this existed, but I stumbled on it yesterday.
When trying to use the Sync theme media option, it takes a couple seconds until I receive "Unable to locate TV Tunes in Kodi".

I installed this version:
https://github.com/latts9923/service.tvtunes

This seems to be the currently maintained version. However, even after installing, I am still receiving the same message.

### Steps to Reproduce

1. Open Jellyfin addon in Kodi
2. Manage Libraries
3. Sync theme media

### Operating System

Xbox One (Windows 10.0.26100.7010)

### Jellyfin Server Version

10.11

### Kodi Version

21.2

### Addon Mode

Addon-mode

### Logs

[j4k-tvtunes.log](https://github.com/user-attachments/files/26169471/j4k-tvtunes.log)

### Server Logs

_No response_

### Additional Information

Looking for where that message is emitted, it seems to be here:

https://github.com/jellyfin/jellyfin-kodi/blob/02dd16ee01035ba02f4f89809e236a7880efa42d/jellyfin_kodi/entrypoint/default.py#L1162-L1171

This checks for availability of `script.tvtunes`, however the addon is now a service, so it would have to be `service.tvtunes`. I don't know whether this would require any more changes though. The [addon thread](https://forum.kodi.tv/showthread.php?tid=347428&pid=2887078) on the Kodi forums says "The only difference is that the add-on is now a service instead of a script.", but that was also a while ago, so not sure if that still holds true.

ErwinBr:
The **Sync Theme Media** option in the Jellyfin-Kodi add-on menu doesn't work anymore because it refers to a version of TVTunes that is no longer maintained (and not python 3 compatible). The new maintainer updated the add-on and made it a service instead of a script.

This PR fixes this by:

1. Referring to service.tvtunes
2. Updating the folder creation part to ensure that the path (library + media name) is built using strings instead of a mix between strings and bytes, which threw an error.
3. Updating tvtunes_nfo in xmls.py to refactor the iteration method.

This PR requires further testing. I am only able to confirm that the custom paths created by Sync Theme Media menu option appear logical, **but I don't know if TVTunes is picking it up correctly**. I do also confirm that the NFO files created inside those folders are linking to a correct path (URL) on my Jellyfin server, and that this path leads to an audio file (the theme music) that can be played in the browser). It's looking promising.

Fixes: https://github.com/jellyfin/jellyfin-kodi/issues/1113

### DIFF
--- a/jellyfin_kodi/entrypoint/default.py
+++ b/jellyfin_kodi/entrypoint/default.py
@@ -1159,9 +1159,9 @@ def get_themes(api_client):
     if not xbmcvfs.exists(library + "/"):
         xbmcvfs.mkdir(library)
 
-    if xbmc.getCondVisibility("System.HasAddon(script.tvtunes)"):
+    if xbmc.getCondVisibility("System.HasAddon(service.tvtunes)"):
 
-        tvtunes = xbmcaddon.Addon(id="script.tvtunes")
+        tvtunes = xbmcaddon.Addon(id="service.tvtunes")
         tvtunes.setSetting("custom_path_enable", "true")
         tvtunes.setSetting("custom_path", library)
         LOG.info("TV Tunes custom path is enabled and set.")
--- a/jellyfin_kodi/entrypoint/default.py
+++ b/jellyfin_kodi/entrypoint/default.py
@@ -1197,9 +1197,12 @@ def get_themes(api_client):
             items[item["Id"]] = folder
 
     for item in items:
+        folder_name = items[item]
+        if isinstance(folder_name, bytes):
+            folder_name = folder_name.decode("utf-8")
 
-        nfo_path = os.path.join(library, items[item])
-        nfo_file = os.path.join(nfo_path, "tvtunes.nfo")
+        nfo_path = os.path.join(str(library), folder_name)
+        nfo_file = os.path.join(str(nfo_path), "tvtunes.nfo")
 
         if not xbmcvfs.exists(nfo_path):
             xbmcvfs.mkdir(nfo_path)
--- a/jellyfin_kodi/helper/xmls.py
+++ b/jellyfin_kodi/helper/xmls.py
@@ -26,7 +26,7 @@ def tvtunes_nfo(path, urls):
     except Exception:
         xml = etree.Element("tvtunes")
 
-    for elem in xml.getiterator("tvtunes"):
+    for elem in xml.iter("tvtunes"):
         for file in list(elem):
             elem.remove(file)
 


oddstr13:
Looks like there is some info about the addon here https://github.com/latts9923/service.tvtunes/wiki

I see that you also have an issue open at https://github.com/latts9923/service.tvtunes/issues/2

It could be that this addon only worked with "native mode", or that it relied on the online parts mentioned that have been removed.


Depending on whether you can figure this out and get it working well with "addon mode" or not, it might make more sense to remove support for the addon.
And if we wish to keep the functionality we might have to implement it anew. If so, we should research whether there is native Kodi support for playing theme music on the menu and see if we can piggy back into that.

ErwinBr:
> Depending on whether you can figure this out and get it working well with "addon mode" or not, it might make more sense to remove support for the addon. And if we wish to keep the functionality we might have to implement it anew. If so, we should research whether there is native Kodi support for playing theme music on the menu and see if we can piggy back into that.

The plugin's not actively maintained and the requester on our side didn't seem very much interested in it anymore either, so I dropped looking into it further. The PR does fix the logic, and paths to theme songs are resolving, but I couldn't get anything to play in Kodi. It only worked when I put the path in the browser. I believe the issue is now on TV Tunes side. Without their interest, I would indeed suggest to drop the feature altogether.

jellyfin_kodi/entrypoint/default.py
```diff
@@ -1197,9 +1197,12 @@ def get_themes(api_client):
...
 
     for item in items:
+        folder_name = items[item]
+        if isinstance(folder_name, bytes):
+            folder_name = folder_name.decode("utf-8")
 
-        nfo_path = os.path.join(library, items[item])
-        nfo_file = os.path.join(nfo_path, "tvtunes.nfo")
+        nfo_path = os.path.join(str(library), folder_name)
+        nfo_file = os.path.join(str(nfo_path), "tvtunes.nfo")
```
oddstr13, line 1205: why str() and bytes check here?

When will these variables ever be anything but strings, why, and what would that be?

oddstr13:
While this PR might not fix the feature, I also see no real reason not to merge it, except for my questions above.

The xml iter is a legit bug (was removed in 3.9), and if I understand it right, there is no py3-compatable version with the script. id

jellyfin_kodi/entrypoint/default.py
ErwinBr: The folder_name for some reason was causing issues (TypeError's) because it contained bytes. This part could indeed be cleaned up a bit, as it's mostly me trying to isolate and guard the issue I was having. I'll see if I can find a moment to jump back into it.

ErwinBr: Okay, did a test to jog my memory, so what happens if we don't decode the the folder_name is that all the folders created in /tv.kodi.Kodi/data/userdata/addon_data/plugin.video.jellyfin/library will be named starting with b' and ending with ' (e.g. b'Game of Thrones'). This means that TV Shows will not be matched, since it's done on the name. I wasn't sure this happens with all the folders, because initially I did get one tv show back ok (don't ask me why), hence the check for bytes as a precaution. The str casts are indeed unnecessary, i can remove it if it bothers you.

oddstr13: str() on a bytes object would result in the `b"Path"`s.

I'm more wondering how and why they're byte objects at this point in the first place, most strings should be unicode strings already.

What I'm getting at is that this is a bandaid, not a fix for the underlying issue.

PR #1127, opened 2026-04-08
Run the black formatter

oddstr13:


.git-blame-ignore-revs
```diff
@@ -10,3 +10,5 @@ df4994bef23a1f7175d47ff4632084e153d6077f
 0bd2e18c477dc16fce52e229cd38c54069b4f699
 # Tool: black
 091f409f94be9623b74e64ed916b2cd8c6cf8c9c
+# Tool: black
+01bdd435c8fea9e1362d95bf0f4b37238f462139
```
mcarlton00, line 14: Is this just going to keep incrementing until the end of time?  Because it's more than a little dumb, imo

oddstr13: Yep, it's either this or having git blame polluted with non-function-changing commits.

Another alternative would be strictly enforcing the code formatting before merging PRs.

mcarlton00: If you're wanting black formatting to be a requirement, it really needs to get added to CI or a commit hook.  This game of playing catch up later is gonna be a long frustrating battle.

mcarlton00 (approved):

.git-blame-ignore-revs
oddstr13: It is in the linter job, but doesn't run until approved for new contributors

PR #1133, opened 2026-04-19
feat(entrypoint): accept timestamp

Rijul-A:
When another plugin or the user sends a request with a `timestamp=x` parameter, honour it


jellyfin_kodi/entrypoint/default.py
```diff
@@ -104,7 +104,43 @@ def __init__(self):
...
+                        playing_file = player.getPlayingFile()
+                        if settings( "useDirectPaths" ) == "0":
+                            is_current = params["id"] in playing_file
+                        else:
+                            queue = json.loads(window("jellyfin_play.json") or "[]")
+                            for p_item in queue:
+                                if p_item.get("Path") == playing_file:
+                                    if p_item.get("Id") == params["id"]:
+                                        is_current = True
+                                        break
```
mcarlton00: Wouldn't it be more efficient to do something like `item = self.get_file_info(playing_file)` instead of looping through the queue?

mcarlton00 (approved):

PR #1138, opened 2026-05-22
Add support for TV show trailers

bossanova808:
Kodi Nexus (20) is the earliest version of Kodi that JF for Kodi supports.

Kodi Nexus added support for TV show trailers in the Kodi database.   Skins also support these trailers including e.g. the default Estuary and Confluence.

This PR adds support for syncing TV show trailers from Jellyfin to Kodi.

Users would need to re-sync their TV library or libraries to see the trailers.


bossanova808:
Am I supposed to do anything here?  Linting thing is not code I touched?

mcarlton00 (approved):
I have no way to test this, but everything seems like it will work

PR #1140, opened 2026-05-31
Fix segment skipping countdown

Joshless:
The recently introduced segment skipping feature has two main issues:

1. **Trigger delay:** Segment checks currently run inside `report_playback`, which only executes on progress reports (every 10 seconds). Furthermore, the check runs before the player position is updated, causing the skip button to appear very late, or not at all unless manually forcing an OSD update (like pause/resume or seek).
2. **Static countdown:** Once the skip dialog is open, the duration countdown remains static and does not tick down (or every 10 seconds if long enough).

This PR addresses both:
- Spawns a lightweight background thread (`JellyfinSegmentMonitor`) during active video playback that polls and checks segments every 1 second.
- Adds an `update_duration` method to `SkipDialog` to dynamically update the window property and control label.
- Updates `_monitor_skip_dialog` to calculate remaining time and refresh the dialog once per second.

### DIFF
--- a/jellyfin_kodi/dialogs/skip.py
+++ b/jellyfin_kodi/dialogs/skip.py
@@ -123,3 +123,33 @@ class SkipDialog(xbmcgui.WindowXMLDialog):
     def is_cancel(self):
         """Return whether the user cancelled."""
         return self.cancel_requested
+
+    def update_duration(self, remaining):
+        """Update the remaining duration and label properties dynamically."""
+        if remaining < 0:
+            remaining = 0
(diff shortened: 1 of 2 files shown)


mcarlton00:
Isn't this already fixed by #1116?

Joshless:
> Isn't this already fixed by #1116?

Closing for now. #1116 covers the segment timing on master; I’ll test whether the countdown fix still applies cleanly on top of the current build before rebasing/resubmitting.

PR #1145, opened 2026-06-09
Disable black in linting CI for now

mcarlton00:
This needs to be reworked into a proper CI solution.  As it sits now black never runs for the vast majority of contributors and causes linting to fail in code not even touched in the current PR.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -72,6 +72,3 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6.0.2
-      - uses: psf/black@stable
-        with:
-          version: "26.1.0"


oddstr13:
The reason it is failing is because you are either merging failing PRs, or not approving CI to run in the first place (github safety feature for outside contributions).

mcarlton00:
I maintain that if we're going to play this autoformatting game, it needs to happen on commit or be truly automatic in some way.  This endless whack a mole where you only find out there's a problem once you push is just frustrating for everybody involved.  Especially since the rules it tries to enforce directly contradicts what we have defined and have used in tox.ini for years and the editorconfig that got added (that not all editors use).

oddstr13:
https://github.com/jellyfin/jellyfin-kodi/blob/master/.pre-commit-config.yaml#L18

The hooks are prepared, but must be enabled on each repo clone.

The line lenght limit is set to allow black to make the judgement, and for flake8 etc to get out of the way – black is doing the judgement call on whether a line slightly "too long" is fine, or if it's best to line-wrap.

https://test-black.readthedocs.io/en/latest/the_black_code_style.html#line-length

mcarlton00:
Boy, I'm sure glad we have such clear documentation that you need to do manual steps with this third party tool.  After all, it's not like the only indication it exists is in a file that's hidden by default in most editors or a requirements file that 95% of kodi devs will never use or look at because kodi's python env is self contained.

oddstr13:
Well, I briefly debated (with myself) back when I implemented this, to add auto-format on PRs, but decided against it, as resolving conflicts caused by it would be a frustrating experience for people not yet used to git.

PR #1169, opened 2026-07-29
Write playlist and node XML as UTF-8

yorah:
Libraries with non-ASCII characters in their name end up with an empty playlist in Kodi. For example, my `Séries` library showed nothing, while `Films` in the same install worked fine.

`ElementTree.write()` defaults to `us-ascii` without an XML declaration, so the tag is written to the `.xsp` as `S&#233;ries`. Kodi then loads smart playlists through TinyXML's stream parser ([SmartPlayList.cpp#L1236](https://github.com/xbmc/xbmc/blob/Omega/xbmc/playlists/SmartPlayList.cpp#L1236)), and without a UTF-8 declaration TinyXML doesn't decode `&#233;` as a UTF-8 sequence, so nothing matches (when it tries to compare it with the tag in the Kodi video database).

Kodi itself adds a UTF-8 declaration when saving smart playlists ([SmartPlayList.cpp#L1407-L1409](https://github.com/xbmc/xbmc/blob/Omega/xbmc/playlists/SmartPlayList.cpp#L1407-L1409)), so we are doing the same with this change.

(Re) Tested on my CoreELEC 21.3 with a 108-show library: 0 items before, 108 after.

### DIFF
--- a/jellyfin_kodi/views.py
+++ b/jellyfin_kodi/views.py
@@ -281,7 +281,7 @@ class Views(object):
             etree.SubElement(rule, "value").text = view["Tag"]
 
         tree = etree.ElementTree(xml)
-        tree.write(file)
+        tree.write(file, encoding="utf-8")
 
     def add_nodes(self, path, view, mixed=False):
         """Create or update the video node file."""
@@ -355,7 +355,7 @@ class Views(object):
             self.node_all(xml)
 
         tree = etree.ElementTree(xml)
-        tree.write(file)
+        tree.write(file, encoding="utf-8")
 
     def node_root(self, root, index):
         """Create the root element"""
@@ -397,7 +397,7 @@ class Views(object):
         )
 
         tree = etree.ElementTree(xml)
-        tree.write(file)
+        tree.write(file, encoding="utf-8")
 
     def node(self, folder, view):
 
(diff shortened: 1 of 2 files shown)


Rijul-A (approved):

yorah:
Hello @oddstr13 ,
This is a real issue, and I've been running this fix on my box since I opened the PR.

Open to fixing it another way if you think there is a better way.

I also read the LLM policy you included in the README, and I understand what this PR does. The PR body is not just LLM output, but reviewed and tweaked manually by me. If you prefer, I can use my own words for everything, but I believe it won't be clearer. Tell me if you want me to do that.

oddstr13:
@yorah PR body looks like pure LLM puke, so just based on that I should nuke this PR.

The changes themselves do look mostly reasonable, but do nuke the slop-test.

Please also add `xml_declaration=True`, so that the xml encoding hint is also written.

Do those things, and I'll overlook it this time.

Preferrably also update the PR description to include links to documentation or precedence showing that this is what Kodi expects.

yorah:
@oddstr13 thanks for your feedback, I apologize for the LLM-heavy PR description (as said, I did review it and modify it before posting, but that was indeed too verbose).

I rewrote it, with links to Kodi (especially how Kodi is writing playlists itself). Tell me if it's better.

Regarding the actual fix, I added `xml_declaration=True` (good call), and dropped the tests.

Squashed and rebased on master.

oddstr13 (approved):

yorah:
Thanks!❤️

PR #1181, opened 2026-08-24
feat(playutils): include descriptive title in subtitle filenames

roosnic1:
Subtitle stream filenames now include a sanitized version of the stream's title or display title, along with the stream index, in addition to the existing source ID, language, and codec. This makes it easier to distinguish between multiple subtitle tracks with the same language when selecting subtitles in the player.

### DIFF
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -4,6 +4,7 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 #################################################################################################
 
 import os
+import re
 from uuid import uuid4
 
 import requests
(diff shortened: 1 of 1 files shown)


Rijul-A (approved):
Tested this one out, thanks!

oddstr13:
What is the purpose of the regex here?
It is, as far as I can tell, filtering out anything that isn't ascii alphanumeric.

Give me examples of expected input and output?

Will break completely for any non-latin scripts.

roosnic1:
@oddstr13 thank you for the comment.

The purpose of the regex is to make sure that there are no problematic characters from the subtitle `title` or `display_title` in the filename. But you're right and i didn't think about non-Latin scripts. I changed the regex to just exclude the problematic characters in the commit [40cb53b](https://github.com/jellyfin/jellyfin-kodi/pull/1181/commits/40cb53bd3828dd3ee3a20828208519a0f79c2a80).

| expected input  | expected output |
| ------------- | ------------- |
| deutsch-benutzer:in | deutsch-benutzerin  |

I hope this makes more sense now.

oddstr13:
Ah, windows filename sanetization?

oddstr13:
I'm putting this PR on hold, as there are more considerations such as filename length and pathname length restrictions that we need to ensure that we adhere to on all platforms and filesystems.

More details on windows file names can be found here: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

os.pathconf is relevant for getting permitted length on unix-like systems.

We might also consider unicodedata to aid in slugifying international characters if needed.

PR #1187, opened 2026-09-01
ci: update release-drafter to v7.7.0

Rijul-A:
The upgrade of the action has the following changes

- auto labeling of PRs is now mandatory, handled through a custom action inherited from `jellyfin/jellyfin-meta-plugins`
- 2 feature deprecations (but not removal): the version-resolver and the categories labels in meta-plugins. The first one is handled by extending from `jellyfin/jellyfin-meta-plugins` and applying the `when` fields. The other is left for that repository, since we cannot modify it here.

Supersedes #1105 

Observe the massive note at the top of `.github/workflows/autolabeler.yaml`; it's left there as a warning.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,20 +1,27 @@
-_extends: jellyfin/jellyfin-meta-plugins
+_extends:
+  from: jellyfin/jellyfin-meta-plugins
+  strategy:
+    categories: append
 
 name-template: "Release $RESOLVED_VERSION"
 tag-template: "v$RESOLVED_VERSION"
 version-template: "$MAJOR.$MINOR.$PATCH"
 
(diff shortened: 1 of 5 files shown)


oddstr13 (changes requested):
I think it would make sense to move away from inheriting from the `jellyfin/jellyfin-meta-plugins`, that repo is aimed at Jellyfin server plugins, and those have different release processes than the Kodi addons.

I'd like more input (second opinion) from other team members before making this decision tho, please stand by.

oddstr13:
The feedback I've gotten affirms my choice of moving release-drafter config and workflows fully into this repo, it is used in a unique way as part of the release process here, and it makes sense to have everything tunable locally.

PR #1188, opened 2026-09-01
Expand deny list of files in build zip

Rijul-A:
- Exclude all .env* files, since they may contain secrets
- Exclude *.log files which may have sensitive information
- Drop .tox and .nox which have isolated test-session envs and artefacts
  which are machine specific
- Drop OS specific files
- Lastly, exclude the local Python environment .venv

### DIFF
--- a/build.py
+++ b/build.py
@@ -91,6 +91,10 @@ def file_filter(file_name: str) -> bool:
         and not file_name.endswith(".pyo")
         and not file_name.endswith(".pyc")
         and not file_name.endswith(".pyd")
+        and not file_name.startswith(".env")
+        and not file_name.endswith(".log")
+        and file_name != ".DS_Store"
+        and file_name != "Thumbs.db"
     )
 
 
@@ -106,6 +110,9 @@ def folder_filter(folder_name: str) -> bool:
         ".mypy_cache",
         ".pytest_cache",
         "__pycache__",
+        ".venv",
+        ".tox",
+        ".nox",
     ]
     for f in filters:
         if f in folder_name.split(os.path.sep):


oddstr13:
I'd like some context on why/when the dep became unused.

The gitignore addition stands well on its own (separate PR), but should simply be *.zip unless there are good reasons not to (exceptions can be added in the future if required).

What is the reasoning behind inverting the filtration logic here?
The build script is primarily used in the CI pipelines, where we are working from a clean build dir.

Rijul-A:
> I'd like some context on why/when the dep became unused.

I will check and respond.

> The gitignore addition stands well on its own (separate PR), but should simply be *.zip unless there are good reasons not to (exceptions can be added in the future if required).

See #1190, I will remove the change from this PR and squash.

> What is the reasoning behind inverting the filtration logic here? The build script is primarily used in the CI pipelines, where we are working from a clean build dir.

I have been doing a lot of testing and development outside of CI where I typically do not have a clean build directory. Using an allow list prevents unrelated local files from ending up in the build and makes local iteration easier.

That said, given that the script is primarily intended for CI and runs from a clean directory there, I don't think the local-development benefit alone is a strong reason to change the existing behavior. I'm happy to revert this and keep the PR focused on the unused dependency.

Rijul-A:
> > I'd like some context on why/when the dep became unused.
> 
> I will check and respond.
> 

The dependency was added in d77d619b `Add support for upnext addon` by @mcarlton00, however, it was never imported into the Python code. Looking at the [`service.upnext` wiki](https://github.com/im85288/service.upnext/wiki/Integration) we can see that integration can be done either with the external add-on or via `NotifyAll` and `onNotification`; the code base uses the latter approach and has done so since the first instance of integration.

oddstr13:
Looks like the signals dep was added with the intent of using it then, but the implementation without it was chosen instead.

Split that dep cleanup out to its own PR, as that's an easy merge then – I'm not sure of whether I want the rest of the changes here yet.

oddstr13:
I think it would be more interesting to know which files you encounter that aren't supposed to go into the release file, it might make more sense to add them to the filter depending on what it is.

Rijul-A:
> I think it would be more interesting to know which files you encounter that aren't supposed to go into the release file, it might make more sense to add them to the filter depending on what it is.

The concrete ones I had in mind are local development artifacts such as `.venv/`, `.idea/`, coverage/test artifacts (`.coverage*`, `.tox/`, `.nox/`, `htmlcov/`), `.DS_Store`/`Thumbs.db`, `*.log`, and local files such as `.env`. Most of these are already in `.gitignore` but aren't covered by the current build filters, so they can be picked up when running the build from a non-clean working tree. Personally I hit the `.venv` and `.env` most often.

`__pycache__`, `.pytest_cache`, `.mypy_cache`, and `*.py[cod]` are already covered by the build filters.

I don't think blindly mirroring `.gitignore` is right either; for example, `addon.xml` is ignored because it's generated but is intentionally part of the release.

oddstr13:
.env files shouldn't be included for sure, same with the .venv dir.
*.log files have a high probability of containing somewhat sensitive info, and should def. be excluded too.

I think I opted specifically not to add htmlcov to the list in case it might be useful to include in some test release, but no strong opinion there.

.idea sounds like .vscode, which I explicitly chose to keep included (even tho they are editor config files, I do consider them kind of part of the source, even if not really useful 99% of the time) - not a particularly strong opinion on those tho.

What does .tox and .nox contain? presumably similar to .mypy_cache ?

mac and windows thumbnail/meta files tend to sneak in everywhere, we should probably ignore those too.

keep in mind that the .gitignore originate from a template, and as such contain a fair bit of defaults not strictly needed for this specific project.


In all, I'd be more comfortable keeping the current structure, but adding some more ignores.
Adding a new file and not finding it in the build is more likely to cause confusion for a while if one isn't aware of (or remember) the explicit include list.

oddstr13 (approved):

PR #1189, opened 2026-09-02
Use correct key for log level

Rijul-A:
Within this file, the `log_level` key in the `settings` dictionary is used only to notify log level changes. However, the new value of the log level was being cached under the wrong key, resulting in the value deviating from the right one upon storage.

It is a fairly minor issue and would be visible only when fiddling with the log level settings repeatedly. I noticed it while working on one of my other pull requests.


oddstr13 (approved):
Good catch

PR #1179, opened 2026-08-18
feat: add artwork to media segment dialogs

Rijul-A:
Add an image to the `Skip Segment` button. The image is chosen in the following priority order

1. The trickplay image (if available on the server) immediately after the end of the segment, or the next episode thumbnail if the segment type is Credits
1. Current episode thumbnail
1. Season landscape
1. TV show landscape
1. Season poster
1. TV show poster
1. TV show fanart

The trickplay feature is gated by a boolean setting to prevent unexpected network activity. It defaults to false.

The trickplay artwork is extracted by Pillow (new dependency) from the sprite sheet and cached for 7 days or 64 MB in Kodi's temporary directory. Tests are included.

[Screenshots](https://github.com/jellyfin/jellyfin-kodi/pull/1179#issuecomment-5409593719)

PR #1191, opened 2026-09-02
Add image to segment dialog

Rijul-A:
Add artwork to skip segment buttons, if available

Expand the size of the segment dialog to fit a landscape image just
above the buttons. The image's original aspect ratio is retained to
avoid distortion tied to the size.

Since segments can be added to any media type, use multiple fallback
options to (attempt to) guarantee availability. If no image exists,
collapse the newly added image placeholder back to the traditional
button-only UI

Base case, when a 16:9 image is available: 
 
image

&nbsp; 

`<aspectratio>keep</aspectratio>` preserves the source aspect ratio, so portrait artwork (if any) is displayed without distortion: 

image

&nbsp;

When no eligible artwork is available, the dialog collapses to the existing button-only layout: 
 
image

&nbsp;
 
The unused space on the right side of the button-only layout is also present in master and is unchanged by this PR. The buttons use auto widths with minimum sizes, leaving room for longer labels.

Partially supersedes #1179


jellyfin_kodi/player.py
```diff
@@ -722,7 +726,10 @@ def _show_skip_button(self, segment_type, duration, end_time):
                 "default",
                 "1080i",
             )
-            self.skip_dialog.set_skip_info(segment_type, duration)
+            segment_image = None
+            if segment_type == "Credits":
```
oddstr13: Worth noting that credits might not be the end of the episode, many things have post-credit snippets, and quite a few of them are not plain next-episode-preview.

jellyfin_kodi/dialogs/skip.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, *args, **kwargs):
 
         xbmcgui.WindowXMLDialog.__init__(self, *args)
 
-    def set_skip_info(self, segment_type, duration):
+    def set_skip_info(self, segment_type, duration, segment_image):
```
oddstr13:
```suggestion
    def set_skip_info(self, segment_type, duration, segment_image=""):
```
Seeing as we default to empty string further down, I think making it optional here makes sense.

jellyfin_kodi/player.py
Rijul-A: Good point. Would using the current item's artwork for Credits, like the other segment types, make more sense here rather than trying to infer whether the credits are actually at the end of the episode?

Rijul-A: FWIW I just followed what the `upnext` addon does. We even send an event to it for Credits specifically with this very image.

oddstr13: I have not used segment skipping in Jellyfin myself, so I don't know what the best UX ergonomics here are.
I don't have my library populated with the data, and have not tested this functionality myself.

I just know that some people would want to skip the long end credits to get to the exclusive video snippet which may or may not even be at the end of the credits. (Marvel movies are famous for this, and some anime have a mini-chibi show at the end for example, both which may contain lore not otherwise available).

If you think the current implementation is suitable even after this info, we'll just go with it, we can always adjust later.

Rijul-A: You are right because UpNext always lands you at the next item while skipping a Credits segment may or may not land you there. Removed this feature.

oddstr13:
No longer contains any Python changes, was that intentional?

Have you tested the latest changeset to be working?

Rijul-A:
> No longer contains any Python changes, was that intentional?

Yes, it's not needed if we don't have the exception for next episode artwork.

> Have you tested the latest changeset to be working?

Will check in a bit.

Rijul-A:
Recap example for a show
image

&nbsp; 

Intro example for a show
image

&nbsp; 

Outro example for a movie, to demonstrate that the image shows up for non-TV shows too
image

Preview
image

&nbsp; 

No image

image

&nbsp; 

Observe that the text is scrolling if it overflows; this applies even when the button is not in focus.

oddstr13:
I would prefer to avoid scrolling if possible.
I would also prefer the close button to be aligned to the image, and for the margins to be consistent.

@erikbucik any additional input here? Note that this is part of the Kodi UI, not your normal Jellyfun UIs.

Rijul-A:
> I would prefer to avoid scrolling if possible. I would also prefer the close button to be aligned to the image, and for the margins to be consistent.
> 
> @erikbucik any additional input here? Note that this is part of the Kodi UI, not your normal Jellyfun UIs.

We can do that by setting a constant width for both of the buttons. I did that in a version of this PR but undid it because `master` has a variable width currently.

https://github.com/jellyfin/jellyfin-kodi/blob/51e3b7a7799aa6250ae422a10d5e1529781d5d42/resources/skins/default/1080i/script-jellyfin-skip.xml#L53

https://github.com/jellyfin/jellyfin-kodi/blob/51e3b7a7799aa6250ae422a10d5e1529781d5d42/resources/skins/default/1080i/script-jellyfin-skip.xml#L69

If we remove the `<scroll>` tag though, the text is clipped from both sides since the button's text is currently center aligned. Kodi doesn't have an ellipsis overflow option as far as I can tell.


EDIT: Some more facts about the buttons

```
<height>56</height>
<font>font13</font>
<aligny>center</aligny>
<align>center</align>
```

`textoffsetx` adds horizontal padding, leaving slightly smaller usable text width than the button width
   - Skip: 20
   - Close: 15
- The controls are designed at the addon’s `1080i` resolution; Kodi scales them on other output resolutions/UI scales.
 - Localized strings are the main risk: English `Close` fits easily, but a longer translated close label or a long segment/duration label may crop.

Options are
1. Lower the font size within the text
2. Align (horizontally) to the left side so that the `Skip` part always shows
3. Reduce the `textoffsetx`

oddstr13:
I was more thinking overall wider box to make scoll less likely, maybe move the type and time info out of the button to a separate line?

It is a very long time since I did any actual UI work in this addon, and the last time was just adjusting the settings dialogs a tiny bit, so I'm not much help with the particulars of skinning unfortunately.

Rijul-A:
image

Something like this?

resources/language/resource.language.en_gb/strings.po
```diff
@@ -1261,10 +1261,6 @@ msgctxt "#33262"
 msgid "Skip {0} ({1})"
 msgstr "Skip {0} ({1})"
 
-msgctxt "#33263"
-msgid "Close"
-msgstr "Close"
```
Rijul-A: Replaced by https://github.com/xbmc/xbmc/blob/84b1acbc2b5af66154992506af0038010a14ace7/addons/resource.language.en_gb/resources/strings.po#L9005; carved out in #1202

oddstr13:
I was thinking the text would go between the image and button, but that also works.

Much cleaner than having the button scroll either way

erikbucik:
> @erikbucik any additional input here?

###### _I don't use Kodi - so I don't know what the rest of the UI looks like. I assume this is placed in a corner (bottom-right)?_

#### My main question is  **Why... do you need the image?**
What information do you understand faster (or slower) when seeing the image?
Does it really help you make the decision faster? Or is it just an additional point that you need to process when making the decision.

###### .

_afaik **that image is a random frame from somewhere around the time stamp**. So if you decide to skip, that's what you'll see?_

`↪` The same applies to chapters:
The way we get chapter images is: a program extracts a frame from a specified time in the video. That frame has absolutely no context to anyone unless they have (_made it_) already seen the video (and even then it's pretty hard to recognise).
Video chapter images would work if each image was designed specifically to <del>represent</del> that chapter. _Like chapter titles are in books._

---
The last style looks best (no scrolling/cut off text). 👍

erikbucik:
Side note: The other button says "Close". If you click it, does it 'close' the video player?

The whole card/popover/dialog is a yes/no question, isn't it?
Therefore the title should be phrased like a question:
```
Skip intro?
- Skip
- Cancel
```
Or maybe it's just an info card:
```
"This is and intro."
- Skip
- Watch
```
What are some other labels for the 'no' button...
- Continue
- Hide (no, that doesn't work)
- ...

_The segment length is best placed in the 'title'._

oddstr13:
I guess "Dismiss" would be an appropriate word for the cancel button?

Rijul-A:
> > @erikbucik any additional input here?
> 
> ###### _I don't use Kodi - so I don't know what the rest of the UI looks like. I assume this is placed in a corner (bottom-right)?_

Yes, it's on the bottom right.
image


> #### My main question is **Why... do you need the image?**
> 
> What information do you understand faster (or slower) when seeing the image? Does it really help you make the decision faster? Or is it just an additional point that you need to process when making the decision.
> ###### .
> 
> _afaik **that image is a random frame from somewhere around the time stamp**. So if you decide to skip, that's what you'll see?_
> 
> `↪` The same applies to chapters: The way we get chapter images is: a program extracts a frame from a specified time in the video. That frame has absolutely no context to anyone unless they have (_made it_) already seen the video (and even then it's pretty hard to recognise). Video chapter images would work if each image was designed specifically to ~represent~ that chapter. _Like chapter titles are in books._

Right now there is no correlation between the timestamp and the image.

My next PR would be to replace this static image with an image extracted from trickplay. I attempted that in #1179 but decided to split that into 2 to make it easier to review.

IMO that would make it much more useful for binge watching, especially for shows that start with an Intro so you know from the image whether you'll skip this entire episode.

To be fair, even without any utility, it makes the skip prompt feel like an intentional media overlay rather than a transient button.
 
> The last style looks best (no scrolling/cut off text). 👍

Thanks.

> I guess "Dismiss" would be an appropriate word for the cancel button?

Makes sense to me.

oddstr13:
I know I asked you to rebase and force push earlier, but once you have gotten in-depth reviews and haven't been explicitly asked to rebase or clean out files from history, best would be to make new commits so that it is easier to follow which changes are new.

There is no need to keep the branch up to date with master with merges or rebases, unless there is a merge conflict or significant interaction with recently merged code (one PR depending on another).

When I want to try out the code on top of master, doing the rebase locally before starting Kodi is usually just a single command :smiley_cat:

---

This PR is probably ready as-is, but I would like for an independent test before merge.
I don't have segments set up in my library right now, so it might take me a bit before I get around to set up a test-server for that.
If anyone else are able to test and give feedback before I get around to that, it'd be appreciated!

Rijul-A:
image

This is from 603428fd6450afad57d477b78e08fb36935f693e. The seekbar on the top doesn't show up automatically; I was scrolling through to get the preview to show up.

Rijul-A:
Screenshots from 9e52d9fd03a6a7d97fb4620d8577fe4b7a4e3c01

The current episode artwork shows up first now, which is the Jellyfin server's Primary as defined in `actions.py`. The quality badge is the artifact of a server level add-on I have installed.

screenshot

&nbsp;

For a movie, the Primary is the portrait artwork of the movie in the server code. The dialog handles portrait without distortion.

screenshot2

&nbsp;

Even though I have traced the generation process and the fallback process completely now, I would also prefer to wait for some user feedback before merging this.

jellyfin_kodi/dialogs/skip.py
```diff
@@ -61,32 +76,18 @@ def set_skip_info(self, segment_type, duration):
...
-        button_label = translate(33262).format(segment_label, duration_text)
+        skip_detail = "{} • {}".format(segment_label, duration_text)
 
-        # Use setProperty so it's available to the skin
-        self.setProperty("skip_label", button_label)
+        # Use setProperty so detail text is available to the skin.
+        self.setProperty("skip_detail", skip_detail)
         self.setProperty("segment_type", segment_type or "")
         self.setProperty("duration", duration_text)
+        self.setProperty("segment_image", get_player_art())
```
oddstr13, line 85: Would this override existing segment images if this is properly populated somewhere?

jellyfin_kodi/dialogs/skip.py
```diff
@@ -61,32 +76,18 @@ def set_skip_info(self, segment_type, duration):
...
         LOG.debug(
-            "SkipDialog: set_skip_info segment=%s, label=%s", segment_type, button_label
+            "SkipDialog: set_skip_info segment=%s, detail=%s", segment_type, skip_detail
         )
 
-    def onInit(self):
-        """Initialize the dialog controls."""
-        LOG.debug("SkipDialog.onInit called")
-
-        # Try to set button label directly as well
```
oddstr13, line 80: I'm wondering why this function was needed, got some insights informing why you're removing it now?

resources/skins/default/1080i/script-jellyfin-skip.xml
```diff
@@ -35,52 +37,111 @@
...
+						<texturefocus colordiffuse="FF00A4DC">white.png</texturefocus>
+						<texturenofocus colordiffuse="33FFFFFF">white.png</texturenofocus>
+						<pulseonselect>no</pulseonselect>
 					</control>
 				</control>
 			</control>
 		</control>
 	</controls>
-</window>
+</window>
```
oddstr13:
```suggestion
</window>

```

Missing final newline – editorconfig should be insisting on this, but it looks like we're not checking that in CI (yet)

jellyfin_kodi/dialogs/skip.py
Rijul-A: Yes it would do that. In Kodi source code it's a dictionary.
```c
m_mapProperties[strKey] = value;
```

jellyfin_kodi/dialogs/skip.py
Rijul-A: It was originally added in #1089 (`6a2c1264`). The initial implementation set the dynamic skip label directly from Python; `b0aa205` then changed that to expose the value through `Window.Property(skip_label)` for use by the XML, while keeping the direct `setLabel()` in `onInit()` as an additional fallback.

With this PR, the button label itself is no longer dynamic, so there is no longer a Python-generated button label for `onInit()` to apply.

There is still dynamic text in the dialog, e.g. `Introduction • 1m 30s`, but that's a separate label populated through `Window.Property(skip_detail)` and consumed by the XML. That property-to-XML path is the intended mechanism for that value; adding another `getControl(...).setLabel(...)` path for `skip_detail` would just duplicate the same state in Python without addressing a known failure of the property mechanism.

So the `onInit()` fallback being removed here is specific to the old dynamic button-label implementation from #1089, rather than something required for every dynamic label in the dialog.

Rijul-A: I also checked Kodi’s Git history. Dynamic button labels have been supported as an info-label path for years; I found no corresponding fix or compatibility note requiring an additional imperative `setLabel()` call after XML property binding.

jellyfin_kodi/dialogs/skip.py
oddstr13: The question here being whether we already sync segment images from Jellyfin or if that is something we'll do in the future, and whether this function would then be in the way of that?

If it would be a problem but isn't currently, I think a TODO comment in the function would be appropriate

Rijul-A: Segments don't have images on the server; `MediaSegmentDto` under v12.0 (and since introduction I believe) contains only `Id`, `ItemId`, `Type`, `StartTicks`, and `EndTicks`.

Are you referring to some other sorts of images? Or maybe I am misunderstanding you.

oddstr13: Jellyfin has the following scheduled tasks which are somewhat related, and based on those, I would expect potential availability of suitable images somewhere server-side

- Extract Chapter Images
- Generate Trickplay Images
- Keyframe Extractor
- Media Segment Scan

Chapter Images sound like exactly what I'm thinking of, but it's possible that's only for mkv chapters and not the segment api, but I thought the former scanned into the latter?

Trickplay Images may not match exactly with segments, but the could be useable (they're used to show preview while hovering/seeking in jf-web afaik).

I do believe these are 10.11 maybe even 10.10 features

oddstr13: These are the type of images I was thinking was in play here, not the primary image of the item (those only make sense for skipping to next item in the series/list)

Rijul-A: I checked the related server-side image/timing data:

| Task | Output | Relationship to `MediaSegment` |
| --- | --- | --- |
| **Extract Chapter Images** | `ChapterInfo` records containing `StartPositionTicks` and `ImagePath` | Separate chapter records. There is no `MediaSegmentId` or automatic chapter → segment association. A chapter and segment can overlap in time, but Jellyfin does not map them. |
| **Generate Trickplay Images** | Scrubbing-preview tiles plus interval/thumbnail metadata | Associated with the item/media source. There is no segment reference; the appropriate thumbnail is selected from the segment timestamp. |
| **Keyframe Extractor** | Keyframe timestamps (`KeyframeTicks`) | Timing data only; no images and no segment association. |
| **Media Segment Scan** | `MediaSegmentDto` records with `Type`, `StartTicks`, `EndTicks`, etc. | Creates the actual segments, but there is currently no image/artwork field associated with a segment. |

So as far as I can tell, there is currently no server-provided segment image for this property to override.

That said, Trickplay is a better source than the item's primary artwork for this. I plan to handle that separately by deriving the displayed image from the segment timestamp using Trickplay; see #1179 for the intended shape of that.

This PR is intentionally self-contained: the current item-art fallback is useful on its own, and this can be released independently even if the Trickplay follow-up is not merged.

oddstr13: I may be remembering discussion around media segments and chapters that didn't end up getting implemented then, because there was discussion about them being essentially the same thing (just different sources)

oddstr13 (changes requested):
I consider this PR mostly ready for merge, only one change request and some independent testing needed.

Add `Dismiss` as a translation string, and use it in place of `Cancel`.


A toggle for enabling/disabling the image would likely be desirable, but do that in a new PR.

PR #1185, opened 2026-08-26
Avoid errors at upgrade/uninstall

Rijul-A:
When Kodi is tearing down the add-on, `xbmcaddon.Addon` is not available. Relying on it without wrapping it in a `RuntimeError` reports an exception in the logs and causes a notification to appear with the complaint. Wrap only some instances of it very narrowly to avoid such issues. Note that this prevents RuntimeError traces from this add-on in the logs, as well as the notifications; however, it does not prevent Kodi's own logs of the format `EXCEPTION: Unknown addon id`.

Based on #1163
Fixes #324 

Example logs, with this patch added, when re-installing the add-on over itself
```
2026-09-02 14:54:57.266 T:3560587   error <general>: EXCEPTION: Unknown addon id 'plugin.video.jellyfin'.
2026-09-02 14:54:57.266 T:3560587    info <general>: JELLYFIN.jellyfin_kodi.monitor -> INFO::jellyfin_kodi/monitor.py:387 ---<[ listener ]
2026-09-02 14:54:57.266 T:3560586   error <general>: EXCEPTION: Unknown addon id 'plugin.video.jellyfin'.
2026-09-02 14:54:57.266 T:3560586    info <general>: JELLYFIN.jellyfin_kodi.entrypoint.service -> INFO::jellyfin_kodi/entrypoint/service.py:506 ---<[ EXITING ]
2026-09-02 14:54:57.266 T:3560607   error <general>: EXCEPTION: Unknown addon id 'plugin.video.jellyfin'.
2026-09-02 14:54:57.266 T:3560607    info <general>: JELLYFIN.jellyfin_kodi.library -> INFO::jellyfin_kodi/library.py:111 ---<[ library ]
2026-09-02 14:54:57.266 T:3560586   error <general>: EXCEPTION: Unknown addon id 'plugin.video.jellyfin'.
2026-09-02 14:54:57.266 T:3560586    info <general>: JELLYFIN.jellyfin_kodi.helper.utils -> DEBUG::jellyfin_kodi/helper/utils.py:54 --[ window clear: jellyfin_state ]
2026-09-02 14:54:57.266 T:3560586   error <general>: EXCEPTION: Unknown addon id 'plugin.video.jellyfin'.
2026-09-02 14:54:57.266 T:3560586    info <general>: JELLYFIN.jellyfin_kodi.helper.utils -> DEBUG::jellyfin_kodi/helper/utils.py:54 --[ window clear: jellyfin_serverStatus ]
[...]

PR #1192, opened 2026-09-02
Cache and reload log level

Rijul-A:
Cache the log level instead of retrieving it for every log record. This avoids accessing the add-on settings during teardown, when the add-on may no longer be available.

Reload the cached log level through a dedicated xbmc.Monitor in the logging module when settings change. Keeping this handling local to the logging module also avoids introducing eager imports of loghandler.

The primary advantage of caching the value is that repeated Kodi log lines of the form `error <general>: EXCEPTION: Unknown addon id 'plugin.video.jellyfin'.` are avoided during add-on teardown.

Keep information masking fixed for the lifetime of the logger and add help text indicating that changing the setting requires a Kodi restart.

Follow-up to https://github.com/jellyfin/jellyfin-kodi/pull/1185#issuecomment-5509386365

On the default Kodi skin, the help message for setting `maskInfo`  shows up in the bottom left.
image

### DIFF
--- a/jellyfin_kodi/helper/lazylogger.py
+++ b/jellyfin_kodi/helper/lazylogger.py
@@ -22,6 +22,11 @@ class LazyLogger(object):
             self.__logger = getLogger(self.__logger_name)
         return getattr(self.__logger, name)
 
+    def setJellyfinLevel(self, level):
+        from .loghandler import getHandler
+
+        getHandler().setJellyfinLevel(level)
+
     #####################################################################
     # Following are stubs of methods provided by `logging.Logger`.      #
     # Please ensure any actually functional code is above this comment. #
--- a/jellyfin_kodi/helper/loghandler.py
+++ b/jellyfin_kodi/helper/loghandler.py
@@ -28,6 +28,10 @@ def getLogger(name=None):
     return __LOGGER.getChild(name)
 
 
+def getHandler():
+    return __HANDLER
+
+
 class LogHandler(logging.StreamHandler):
 
     def __init__(self):
@@ -45,13 +49,29 @@ class LogHandler(logging.StreamHandler):
             if server.get("address"):
                 self.sensitive["Server"].append(server["address"].split("://", 1)[-1])
 
-        self.mask_info = settings("maskInfo.bool")
+        try:
+            value = settings("logLevel")
+        except RuntimeError:
+            value = "2"
+        self.setJellyfinLevel(value)
+
+        try:
+            self.mask_info = settings("maskInfo.bool")
+        except RuntimeError:
+            self.mask_info = True
 
         self.level = xbmc.LOGINFO
 
+    @classmethod
+    def setJellyfinLevel(cls, jellyfin_level):
+        try:
+            cls.jellyfin_level = int(jellyfin_level)
+        except (ValueError, TypeError):
+            cls.jellyfin_level = 2
+
     def emit(self, record):
 
-        if self._get_log_level(record.levelno):
+        if self._getLogLevel(record.levelno):
             string = self.format(record)
 
             if self.mask_info:
@@ -68,23 +88,14 @@ class LogHandler(logging.StreamHandler):
             xbmc.log(string, level=self.level)
 
     @classmethod
-    def _get_log_level(cls, level):
-
+    def _getLogLevel(cls, level):
         levels = {
             logging.ERROR: 0,
             logging.WARNING: 0,
             logging.INFO: 1,
             logging.DEBUG: 2,
         }
-        try:
-            log_level = int(settings("logLevel"))
-        except (ValueError, RuntimeError):
-            # If getting settings fail, we probably want debug logging.
-            # ValueError happens if the result is not a number
-            # RuntimeError happens if the addon is being re-installed
-            log_level = 2
-
-        return log_level >= levels[level]
+        return cls.jellyfin_level >= levels[level]
 
 
 class MyFormatter(logging.Formatter):
@@ -127,5 +138,6 @@ __LOGGER = logging.getLogger("JELLYFIN")
 for handler in __LOGGER.handlers:
     __LOGGER.removeHandler(handler)
 
-__LOGGER.addHandler(LogHandler())
+__HANDLER = LogHandler()
+__LOGGER.addHandler(__HANDLER)
 __LOGGER.setLevel(logging.DEBUG)
--- a/jellyfin_kodi/entrypoint/service.py
+++ b/jellyfin_kodi/entrypoint/service.py
@@ -424,6 +424,7 @@ class Service(xbmc.Monitor):
 
             log_level = settings("logLevel")
             self.settings["log_level"] = log_level
+            LOG.setJellyfinLevel(log_level)
             LOG.info("New log level: %s", log_level)
 
         if settings("enableContext.bool") != self.settings["enable_context"]:


oddstr13:
Extreme care needs to be taken when dealing with imports from/in loghandler to avoid import loops.

LazyLogger exists to avoid the consequences of these import loops, by doing late initialization of the logger.

The loghandler.py file is explicitly not imported before the first call to a logger is done, to make sure all other modules the loghandler depend on in any way are done importing and their classes, functions and variables are available for the loghandler.

---

I don't think that it makes sense changing the information masking setting mid-run, as we usually want the whole log, a note about needing a restart to take effect would be more appropriate.

---

When it comes to caching the log level in the LogHandler, the question becomes; how much does the read settings call cost us, and is it worth the added complexity to optimize this away?

If the answer to that question is yes, I think it might be better to handle those events directly in the loghandler.

I see that we are using xbmc.Monitor instances all over the place, so I don't think that creating one more subclass to go with the LogHandler would be a problem. That subclass could then be passed the LogHandler instance created at the bottom of the file (or just access classmethods directly).

Rijul-A:
> Extreme care needs to be taken when dealing with imports from/in loghandler to avoid import loops.
> 
> LazyLogger exists to avoid the consequences of these import loops, by doing late initialization of the logger.
> 
> The loghandler.py file is explicitly not imported before the first call to a logger is done, to make sure all other modules the loghandler depend on in any way are done importing and their classes, functions and variables are available for the loghandler.
> 
> I don't think that it makes sense changing the information masking setting mid-run, as we usually want the whole log, a note about needing a restart to take effect would be more appropriate.
> 
> When it comes to caching the log level in the LogHandler, the question becomes; how much does the read settings call cost us, and is it worth the added complexity to optimize this away?
> 
> If the answer to that question is yes, I think it might be better to handle those events directly in the loghandler.
> 
> I see that we are using xbmc.Monitor instances all over the place, so I don't think that creating one more subclass to go with the LogHandler would be a problem. That subclass could then be passed the LogHandler instance created at the bottom of the file (or just access classmethods directly).


Reworked this based on your feedback.

* Removed the imports/callback handling from `service.py`, so `loghandler` remains late-loaded as intended.
* Added an `xbmc.Monitor` class directly in `loghandler.py` and passed it the instantiated `LogHandler`; it reloads the cached log level through `onSettingsChanged`.
* Left information masking fixed for the lifetime of the logger and added the existing “Restart Kodi to take effect” help text to the setting. This deviates a bit from the PR title so let me know if you'd like it removed.

The primary reason for this change is Kodi's teardown behaviour and not the cost of the settings lookup itself: avoiding the lookup for each log record prevents the repeated Kodi `Unknown addon id` messages once the add-on is no longer available.

Rijul-A:
I benchmarked the settings read in the add-on context. `settings("logLevel")` takes ~444 µs per call on my system. Breaking it down, constructing `xbmcaddon.Addon(...)` takes ~0.39 µs and calling `getSetting()` on a reused Addon instance takes ~0.25 µs, but constructing a new Addon and then calling `getSetting()` takes ~444–448 µs.

The cached integer comparison is ~0.03 µs, so the current per-record settings lookup is roughly 14,800× more expensive than the cached check. The benchmark suggests that avoiding the lookup on every log record is worthwhile from a performance perspective.

Based on the benchmarks, there may be a broader opportunity to reuse the `Addon` object in `settings()`, but that's project-wide behavior and I'd rather leave it out of this PR.

EDIT: I looked into Kodi’s source and found why this happens. Settings are lazily initialized on the first `addon.getSetting()` call for a newly-created `Addon` instance. Subsequent reads on the same object can skip that initialization:

```cpp
if (SettingsInitialized(id) && !bForce)
    return true;
```

So with a reused `Addon` instance, the expensive initialization cost is paid once and amortized over subsequent reads. When Kodi reloads settings for a retained add-on instance, it calls the loader with `bForce = true`, bypassing the early return and refreshing the settings.

jellyfin_kodi/helper/loghandler.py
```diff
@@ -71,24 +72,41 @@ def emit(self, record):
...
+                self.log_level = 2
+
+
+class LogMonitor(xbmc.Monitor):
 
-        return log_level >= levels[level]
+    def __init__(self, handler):
+
+        xbmc.Monitor.__init__(self)
+        self.handler = handler
```
oddstr13:
```suggestion
        self.handler = handler
        xbmc.Monitor.__init__(self)
```
Swap these to avoid potential race condition where SettingsChanged fires before handler is set.
Extremely low probability of it happening, but would be possible I think? Better safe than sorry.

resources/settings.xml
```diff
@@ -682,7 +682,7 @@
 					</constraints>
 					<control type="spinner" format="string"/>
 				</setting>
-				<setting id="maskInfo" type="boolean" label="33164" help="">
+				<setting id="maskInfo" type="boolean" label="33164" help="33110">
```
oddstr13: change belonging to different PR?

oddstr13 (changes requested):
A couple of small things, otherwise looks good – pending functional testing.

I don't think the caching is strictly needed, at only ~20x speedup for that particular part of the logging (if I understood you correctly).

When profiling code like this, you would normally first warm up the cache (run the functions once (or more with different parameters, depending)), and then run the code under test a bunch of times and take the averages. (how many times depend on how quickly the tested code runs, at the execution speeds seen on these functions, that'd be 10k times or more, essentially as many as you can be bothered to wait for – but no need to go completely wild).

resources/settings.xml
Rijul-A: https://github.com/jellyfin/jellyfin-kodi/pull/1192#issuecomment-5515856600

>  -Left information masking fixed for the lifetime of the logger and added the existing “Restart Kodi to take effect” help text to the setting. This deviates a bit from the PR title so let me know if you'd like it removed.

I added it per your "I don't think that it makes sense changing the information masking setting mid-run, as we usually want the whole log, a note about needing a restart to take effect would be more appropriate."

Rijul-A:
> I don't think the caching is strictly needed, at only ~20x speedup for that particular part of the logging (if I understood you correctly).

Thanks; the measured difference between the current per-record settings lookup and the cached comparison is ~14,800×: `settings("logLevel")` takes ~444 µs on my system, while the cached integer comparison takes ~0.03 µs. Of course, that is only the log-level check and not a 14,800× speed-up of logging overall.

One important detail is that warming `settings("logLevel")` does not make subsequent iterations use the same Kodi settings cache: `settings()` creates a fresh `xbmcaddon.Addon` object on every call. Kodi lazily initializes that instance's settings on its first `getSetting()`, so every benchmark iteration intentionally reproduces the same cold-per-instance path used by the logger in `master`.
 
> When profiling code like this, you would normally first warm up the cache (run the functions once (or more with different parameters, depending)), and then run the code under test a bunch of times and take the averages. (how many times depend on how quickly the tested code runs, at the execution speeds seen on these functions, that'd be 10k times or more, essentially as many as you can be bothered to wait for – but no need to go completely wild).

The benchmark did warm each operation before measuring and then used the median of three runs (10k iterations for the slower paths, 1m for the cheap ones).

<details>
<summary>Benchmark code</summary>

```python
# -*- coding: utf-8 -*-
from __future__ import division, absolute_import, print_function, unicode_literals

import logging
import statistics
import time

import xbmc
import xbmcaddon

from jellyfin_kodi.helper.utils import addon_id, settings


ADDON_ID = "plugin.video.jellyfin"
SETTING = "logLevel"

SLOW_READS = 10000
FAST_READS = 1000000
ROUNDS = 3


def benchmark(name, operation, reads):
    samples = []

    # Warm up before measuring.
    operation()

    for _ in range(ROUNDS):
        start = time.perf_counter_ns()

        for _ in range(reads):
            operation()

        samples.append(time.perf_counter_ns() - start)

    median_ns = statistics.median(samples)
    per_read_ns = median_ns / reads

    xbmc.log(
        "[JELLYFIN] %s: median of %d rounds, %d reads: "
        "%.3f ms total, %.3f us/read"
        % (
            name,
            ROUNDS,
            reads,
            median_ns / 1_000_000,
            per_read_ns / 1_000,
        ),
        xbmc.LOGINFO,
    )


def main():
    addon = xbmcaddon.Addon(ADDON_ID)
    log_level = int(addon.getSetting(SETTING))

    levels = {
        logging.ERROR: 0,
        logging.WARNING: 0,
        logging.INFO: 1,
        logging.DEBUG: 2,
    }

    benchmark(
        "addon_id()",
        addon_id,
        SLOW_READS,
    )

    benchmark(
        "xbmcaddon.Addon(literal ID)",
        lambda: xbmcaddon.Addon(ADDON_ID),
        FAST_READS,
    )

    benchmark(
        "xbmcaddon.Addon(addon_id())",
        lambda: xbmcaddon.Addon(addon_id()),
        SLOW_READS,
    )

    benchmark(
        "reused Addon.getSetting(logLevel)",
        lambda: addon.getSetting(SETTING),
        FAST_READS,
    )

    benchmark(
        "reused Addon.getSetting(logLevel) + int",
        lambda: int(addon.getSetting(SETTING)),
        FAST_READS,
    )

    benchmark(
        "new Addon(literal ID) + getSetting(logLevel)",
        lambda: xbmcaddon.Addon(ADDON_ID).getSetting(SETTING),
        SLOW_READS,
    )

    benchmark(
        "new Addon(literal ID) + getSetting(logLevel) + int",
        lambda: int(xbmcaddon.Addon(ADDON_ID).getSetting(SETTING)),
        SLOW_READS,
    )

    benchmark(
        "new Addon(addon_id()) + getSetting(logLevel)",
        lambda: xbmcaddon.Addon(addon_id()).getSetting(SETTING),
        SLOW_READS,
    )

    benchmark(
        "new Addon(addon_id()) + getSetting(logLevel) + int",
        lambda: int(xbmcaddon.Addon(addon_id()).getSetting(SETTING)),
        SLOW_READS,
    )

    benchmark(
        "settings(logLevel)",
        lambda: settings(SETTING),
        SLOW_READS,
    )

    benchmark(
        "cached integer log-level comparison",
        lambda: log_level >= levels[logging.DEBUG],
        FAST_READS,
    )


if __name__ == "__main__":
    main()
```

</details>

oddstr13:
In which case I misunderstood what you where saying in your edit above. ~0.5ms vs <0.1µs would be significant enough to warrant it, especially for debug logging on lower end systems.
Spam-reading the settings file on every log call is not ideal. (I understood your initial benchmarking comment as if this happened only once, not on every log call).

resources/settings.xml
oddstr13: I just noticed that you referenced a translation not touched here, and that you where also touching on translations in several other PRs, all good.

Rijul-A:
If needed, I will share logs from functional testing later.

oddstr13:
Looks good, but I'll want to do an actual test of this PR before merging. (not tonight)

oddstr13:
Log level isn't actually updating when changing the settings with this, could be the monitor is only running in one thread, and that there are more instances?
(I saw at least debug messages from websocket when setting the log level to info)

Rijul-A:
I reproduced this. It doesn't appear to be a limitation on the number of `Monitor` instances; it's thread affinity: the Python thread on which the monitor is created.

`LogMonitor` is created on the root service-script thread. Kodi registers every `xbmc.Monitor()`, but Python monitor callbacks are queued through a callback handler that captures the `PyThreadState` active when the monitor is created. `Monitor.waitForAbort()` calls `MakePendingCalls()`, which is what processes those queued callbacks.

In our startup path, the first:

```python
LOG.info("-->[ service ]")
```

initializes `LogHandler`/`LogMonitor` on the root thread. `ServiceManager` is then started on another thread and the root thread blocks here:

```python
session.start()
session.join()
```

`Service` calls `waitForAbort()` on the worker thread, so callbacks for its monitor are processed there, but that doesn't process the callback queued for the root-thread `LogMonitor`. Consequently `LogMonitor.onSettingsChanged()` never runs and the cached log level remains unchanged.

One way to address this is to have the root service thread continue pumping Kodi pending callbacks while `ServiceManager` runs rather than blocking exclusively in `join()`, using `waitForAbort()`.

I tested this with DEBUG -> INFO -> DEBUG -> DISABLED, and the log level updates without restarting the add-on.

This does require changing the root service loop rather than keeping the change entirely within `loghandler`. I haven't found a way to service the root-thread `LogMonitor` callback while that thread remains blocked in `join()`.

One other avenue that may be worth investigating is whether `settings()` needs to construct a new `xbmcaddon.Addon` instance on every call.

In my benchmark, the current `settings()` path took ~444 µs per read, while `getSetting()` on a reused `Addon` instance took ~0.25 µs. I couldn't find any historical explanation for the per-call construction.

I haven't verified whether retaining an `Addon` instance is safe across profile changes, add-on reloads and teardown, or calls originating from different Python threads, so I don't think that's something we should change based on the benchmark alone. If Kodi's lifecycle and threading semantics permit it, though, it could make per-record log-level reads cheap enough to remove the need for a cached log level entirely.

oddstr13:
I don't think keeping a reference to an initialized Addon thru a reload should be considered safe, regardless of how this actually works in practice.

I think it might be better to go the route via the lazylogger somehow, a custom function on it and do `LOG.setJellyfinLevel()` or some such.

Not sure of how exactly to implement this, but I imagine lazylogger needs a way to get hold of the handler, maybe via a _getHandler() similar to getLogger()

Rijul-A:
> I don't think keeping a reference to an initialized Addon thru a reload should be considered safe, regardless of how this actually works in practice.
> 
> I think it might be better to go the route via the lazylogger somehow, a custom function on it and do `LOG.setJellyfinLevel()` or some such.
> 
> Not sure of how exactly to implement this, but I imagine lazylogger needs a way to get hold of the handler, maybe via a _getHandler() similar to getLogger()

Routed it through `LazyLogger` -> `LogHandler` -> the handler's cached level, as suggested. Tested DEBUG -> INFO -> DEBUG -> DISABLED and it updates correctly without restarting the add-on.

jellyfin_kodi/helper/lazylogger.py
```diff
@@ -23,6 +23,14 @@ def __getattr__(self, name):
             self.__logger = getLogger(self.__logger_name)
         return getattr(self.__logger, name)
 
+    def _getHandler(self):
```
oddstr13: no need for this intermediary, if we end up needing it in the future we'll consider it then, move into the setJellyfinLevel function

tests/test_loghandler.py
```diff
@@ -0,0 +1,32 @@
+import logging
```
oddstr13: Do these tests add any actual value?

tests/test_lazylogger.py
```diff
@@ -0,0 +1,18 @@
+from jellyfin_kodi.helper import LazyLogger
+
+
+def test_set_jellyfin_level(monkeypatch):
```
oddstr13: Does this test add any actual value?

jellyfin_kodi/helper/loghandler.py
```diff
@@ -71,24 +89,16 @@ def emit(self, record):
...
-            # If getting settings fail, we probably want debug logging.
-            # ValueError happens if the result is not a number
-            # RuntimeError happens if the addon is being re-installed
-            log_level = 2
-
-        return log_level >= levels[level]
+    def _get_log_level(self, level):
+        return (
+            self.jellyfin_level
+            >= {
```
oddstr13: easier to read with the levels dict separate

jellyfin_kodi/helper/loghandler.py
```diff
@@ -71,24 +89,16 @@ def emit(self, record):
 
             xbmc.log(string, level=self.level)
 
-    @classmethod
```
oddstr13, line 70: I've not quite decided whether it makes sense to keep this (and set_jellyfin_level) a classmethod.

Can you think of a reason for why we shouldn't? Are we ever likely to want multiple instances of the class with different log levels?

jellyfin_kodi/helper/loghandler.py
```diff
@@ -46,13 +50,27 @@ def __init__(self):
...
+            self.mask_info = settings("maskInfo.bool")
+        except RuntimeError:
+            self.mask_info = True
 
         if kodi_version() > 18:
             self.level = xbmc.LOGINFO
         else:
             self.level = xbmc.LOGNOTICE
 
+    def set_jellyfin_level(self, jellyfin_level):
```
oddstr13: I'd prefer camelCase here, as that is the norm for the library we're extending. Ignore the fact that the private method below doesn't follow this… or convert it too.

oddstr13 (changes requested):
Mostly nits and minor things left.

I don't quite see the value in the included tests tho, do they really exercise things in a useful way?

I could see the value of end-to-end testing of the log paths including filtering and such, but that is more work than appropriate for this PR, and low priority.

tests/test_lazylogger.py
Rijul-A: Just that it flows from LazyLogger to LogHandler correctly. Fairly minor so happy to drop.

tests/test_loghandler.py
Rijul-A: The first one only a little bit; though I do like having the second one there to guard for invalid log levels.

jellyfin_kodi/helper/loghandler.py
Rijul-A: Not in normal use, but I don't see much benefit either way. Perhaps in tests we may want to break the invariant that all handlers share the same level, in which case keeping it as instance state would be useful.

jellyfin_kodi/helper/loghandler.py
Rijul-A: I was on the fence for this one but I chose to follow the private method below instead of the base class. I will change both.

jellyfin_kodi/helper/loghandler.py
oddstr13: Old version didn't really allow for this either, fetching level from settings. Can't think of any tests where we would need two different instances at the same time.

I don't feel super-strong about it, but I was happy being able to make *anything* in this codebase not depend *heavily* on instance state when I first wrote this code 😅 

So much of the codebase is practically untestable without extensive monkey-patching, so when I had the chance I made stuff somewhat stand-alone :P

tests/test_lazylogger.py
Rijul-A: Dropped now that the intermediary function is removed, so it no longer demonstrates a lazy import.

tests/test_loghandler.py
oddstr13: only tests that the function correctly handles invalid input and falls back to debug as far as I can see (second test)

first one would be more useful if it tested that the logger.info() etc call actually pass the message along to kodi when appropriate, but that is best suited for a separate pr

I'd say drop all three tests for now?

jellyfin_kodi/helper/loghandler.py
Rijul-A: Ah, I hadn't considered that the original shape was deliberately preserving one of the few bits of logic that could stay relatively independent of instance state. Then it's probably worth keeping, I will change it back.

Rijul-A: Do you want to make `maskInfo` a class state too?

jellyfin_kodi/helper/loghandler.py
oddstr13: sorry, I should've been clearer; I normally use the following (this codebase is a mess tho)
- camelCase for functions and methods (implies callable)
- snake_case for variables
- PascalCase for classes
- SCREAMING_SNAKE for "static" (often global) variables

mask_info and jellyfin_level should've stayed in that format

Rijul-A: I see; I perfunctorily checked `logging.streamHandler` and it was using camelCase for method variables so I assumed that's what you were looking for. Will bookmark this.

tests/test_loghandler.py
oddstr13: I know the impulse to TESTEVERYTHING!! btw, but right now I'm thinking about the balancing of CI runtime of executing this (currently) 11 times, with the value it provide the project

Rijul-A: Makes sense; I can also see why the matrix is required

This kind of review is useful for me, by the way. I can reason about whether a change is correct, but your experience maintaining this codebase is helping me get a better feel for which tradeoffs are actually worth making here.

jellyfin_kodi/helper/loghandler.py
oddstr13: completely neutral, that one is only set once at instantiation right now, and we didn't really plan on it becoming dynamic at the moment.
probably cleaner to just refer to it using self

oddstr13 (approved):

oddstr13:
Have you runtime-tested the latest change?

Rijul-A:
> Have you runtime-tested the latest change?

Yes, tested from Debug -> Info -> Disabled and it works as expected.

tests/test_loghandler.py
oddstr13: Yep, and I'm also thinking of adding another set of tests with different Kodistubs versions to maybe catch backwards incompatible changes more easily, which would add a few more runs of the tests.
No need to test each Kodi version with each Python version, fortunately, so that would only add a few test runs, not multiply them :upside_down_face:

PR #1195, opened 2026-09-02
Localize segment skip dialog

Rijul-A:
Move the skip-button label, duration formatting, segment labels, and Close button text into the add-on localization resources so the dialog can be translated.

### DIFF
--- a/resources/skins/default/1080i/script-jellyfin-skip.xml
+++ b/resources/skins/default/1080i/script-jellyfin-skip.xml
@@ -64,7 +64,7 @@
 						</control>
 						<!-- Close button -->
 						<control type="button" id="3013">
-							<label>Close</label>
+							<label>$LOCALIZE[33263]</label>
 							<height>56</height>
 							<width min="60">auto</width>
 							<font>font13</font>
--- a/jellyfin_kodi/dialogs/skip.py
+++ b/jellyfin_kodi/dialogs/skip.py
@@ -3,7 +3,7 @@ from __future__ import division, absolute_import, print_function, unicode_litera
 
 import xbmcgui
 
-from ..helper import LazyLogger
+from ..helper import LazyLogger, translate
 
 LOG = LazyLogger(__name__)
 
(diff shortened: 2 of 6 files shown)


oddstr13:
en_gb (just listed as English on weblate) is the source language for translations, it is the only language that should be touched in git to avoid issues, any other translations should be done via weblate (en_us included).

https://translate.jellyfin.org/projects/jellyfin/jellyfin-kodi/

this also goes for your other PRs touching translation strings

resources/skins/default/1080i/script-jellyfin-skip.xml
```diff
@@ -64,7 +64,7 @@
 						</control>
 						<!-- Close button -->
 						<control type="button" id="3013">
-							<label>Close</label>
+							<label>$LOCALIZE[33263]</label>
```
oddstr13, line 67: I don't see this used anywhere else in the addon, link to docs?

oddstr13 (commented):
Looks like this code expects arbitrary text values from the server, with a short list of aliases.
Might be a good idea to see whether the server has a longer list of standard values that also need translation.

resources/skins/default/1080i/script-jellyfin-skip.xml
Rijul-A: https://kodi.wiki/view/Label_Parsing

Rijul-A:
> Looks like this code expects arbitrary text values from the server, with a short list of aliases. Might be a good idea to see whether the server has a longer list of standard values that also need translation.

I checked the server types and the conversion path. The server currently exposes `Intro`, `Outro`, `Recap`, `Preview`, `Commercial` and `Unknown`; Unknown/unrecognized values are dropped by `_convert_media_segments()`. I moved the existing conversion map to a shared constant and added a test asserting that every segment type we accept has a corresponding localized label.

Separately, it looks like the mapping from `Intro` to `Introduction` and from `Outro` to `Credits` doesn't _have_ to exist but I will leave that for another PR. It will allow us to use a single mapping from `SegmentType` to `StringID` and drop the other repeated one.

oddstr13 (approved):

PR #1197, opened 2026-09-03
Remove unsupported Python and Kodi code

Rijul-A:
Since release 1.0.0, this addon has not supported Kodi <= v18 and Python 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@
   <thead>
     <tr>
       <td align="left">
-        :warning: Python 2 deprecation (Kodi 18 Leia and older)
+        :warning: Python 2 unsupported (Kodi 18 Leia and older)
       </td>
     </tr>
   </thead>
@@ -44,12 +44,9 @@
     <tr>
       <td>
         <p>
-          Kodi installs based on Python 2 are no longer supported
-          going forward.
-          <br/>
-          This means that Kodi v18 (Leia) and earlier
-          (Krypton, Jarvis...) is no longer supported,
-          and will cease receiving updates.
+          Python 2 is unsupported. Kodi installs based on
+          Python 2 cannot receive builds or updates. This includes Kodi v18 (Leia)
+          and earlier (Krypton, Jarvis...).
         </p>
         <p>
           Our informal support target is current release±1,
@@ -58,9 +55,6 @@
           Please note that next release is a moving target,
           has a relatively low priority,
           and is unlikely to receive active work before the release candidate stage.
-        </p>
-        <p>
-          The major version of Jellyfin for Kodi will be bumped for the first release without Python 2 support.
         </p>
       </td>
     </tr>
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@
   <thead>
     <tr>
       <td align="left">
-        :warning: Python 2 unsupported (Kodi 18 Leia and older)
+        :warning: Python 2 deprecation (Kodi 18 Leia and older)
       </td>
     </tr>
   </thead>
@@ -44,9 +44,12 @@
     <tr>
       <td>
         <p>
-          Python 2 is unsupported. Kodi installs based on
-          Python 2 cannot receive builds or updates. This includes Kodi v18 (Leia)
-          and earlier (Krypton, Jarvis...).
+          Kodi installs based on Python 2 are no longer supported
+          going forward.
+          <br/>
+          This means that Kodi v18 (Leia) and earlier
+          (Krypton, Jarvis...) is no longer supported,
+          and will cease receiving updates.
         </p>
         <p>
           Our informal support target is current release±1,
@@ -55,6 +58,9 @@
           Please note that next release is a moving target,
           has a relatively low priority,
           and is unlikely to receive active work before the release candidate stage.
+        </p>
+        <p>
+          The major version of Jellyfin for Kodi will be bumped for the first release without Python 2 support.
         </p>
       </td>
     </tr>
--- a/build.py
+++ b/build.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import argparse
 import os
--- a/build.py
+++ b/build.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 import argparse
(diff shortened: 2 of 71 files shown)


jellyfin_kodi/helper/utils.py
```diff
@@ -429,9 +425,9 @@ def convert_to_local(date, timezone=tz.tzlocal()):
         date = parser.parse(date) if isinstance(date, str) else date
         date = date.replace(tzinfo=tz.tzutc())
         date = date.astimezone(timezone)
-        # Bad metadata defaults to date 1-1-1.  Catch it and don't throw errors
-        if date.year < 1900:
-            # FIXME(py2): strftime don't like dates below 1900
```
oddstr13, line 434: IIRC this might've been a issue specifically with some special windows installs (xbox), and I don't think the change you're making here makes much sense. Keeping it as it was should have negligable peformance impact

jellyfin_kodi/objects/kodi/queries.py
```diff
@@ -1,9 +1,8 @@
-from __future__ import division, absolute_import, print_function, unicode_literals
-
-""" Queries for the Kodi database. obj reflect key/value to retrieve from jellyfin items.
-    Some functions require additional information, therefore obj do not always reflect
-    the Kodi database query values.
+"""Queries for the Kodi database. obj reflect key/value to retrieve from jellyfin items.
+Some functions require additional information, therefore obj do not always reflect
+the Kodi database query values.
```
oddstr13, line 3: What is your reason for reformating the docstrings? (both here and elsewhere)

build.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
oddstr13: `sudo apt install python-is-python3`

build.py
```diff
@@ -30,7 +30,7 @@ def indent(elem: ET.Element, level: int = 0) -> None:
             elem.tail = i
 
 
-def create_addon_xml(config: dict, source: str, py_version: str) -> None:
```
oddstr13, line 33: More cleanup is needed around the py2 vs py3 building, leave this file alone for now.

README.md
```diff
@@ -35,7 +35,7 @@
   <thead>
     <tr>
       <td align="left">
-        :warning: Python 2 deprecation (Kodi 18 Leia and older)
+        :warning: Python 2 unsupported (Kodi 18 Leia and older)
```
oddstr13: Leave as is for now, the whole warning should be removed at some point.

jellyfin_kodi/dialogs/usersconnect.py
```diff
@@ -32,7 +30,6 @@ class UsersConnect(xbmcgui.WindowXMLDialog):
 
     def __init__(self, *args, **kwargs):
 
-        self.kodi_version = kodi_version()
```
oddstr13, line 35: Have you verified that this is never used anywhere?

tests/test_helper_utils.py
```diff
@@ -1,15 +1,9 @@
 # -*- coding: utf-8 -*-
-from __future__ import division, absolute_import, print_function, unicode_literals
 
 import sys
 
-# Python 2
-if sys.version_info < (3, 0):
-    zoneinfo = None
-# Python 3.0 - 3.8
```
oddstr13, line 9: Keep the comments – Python 3.8 is still supported, and these comments are a reading aid for the version matching logic. Just add an empty line before the comments to make it more clear.

release.yaml
```diff
@@ -37,12 +37,11 @@ changelog: |-
   + Update release-drafter/release-drafter action to v6.4.0 (#1102) @[renovate[bot]](https://github.com/apps/renovate)
   + Update github/codeql-action action to v4.34.1 (#1099) @[renovate[bot]](https://github.com/apps/renovate)
 dependencies:
-  py3:
```
oddstr13, line 40: See comments on build script.

oddstr13 (changes requested):
All of this was pretty much left alone because it didn't hurt anything during the py2 removal pass.

I'm fairly neutral on most of it, but leave the build script alone, the whole ci-build-publish pipeline needs a rework to clean out the python2 parts, potentially including server-side stuff (I haven't checked).

I also don't see any particular reason for cleaning out the py2 bits from the tests.

The from __future__ imports are noops in versions they are mandatory, so those don't hurt anything, but I guess one can argue they're clutter to be cleaned.
Neutral on those, keep them in tests?

jellyfin_kodi/objects/kodi/queries.py
Rijul-A: `black` did it.

oddstr13: That's new (or old), which version are you using?

Rijul-A: black, 26.5.1

- In upstream/master, `from __future__ import` ... precedes the string, so it is not considered the module docstring. Black preserves it as an ordinary standalone string.
 - In this pull request, removing that import made the string the first module statement. Black recognizes it as the module docstring and normalizes its whitespace/indentation.

jellyfin_kodi/dialogs/usersconnect.py
Rijul-A: Yes; its usage was removed in #263

Rijul-A:
> I also don't see any particular reason for cleaning out the py2 bits from the tests.
> 
> The from **future** imports are noops in versions they are mandatory, so those don't hurt anything, but I guess one can argue they're clutter to be cleaned. Neutral on those, keep them in tests?

That’s fair. The `__future__` imports are harmless on supported Python versions, so I don’t feel strongly about removing them.

The `zoneinfo is None` branch and associated skip only existed for Python 2. They no longer provide executable Python 2 coverage: even before this change, the test suite imported application modules containing Python-3-only syntax such as f-strings, and `test_translations.py` uses `pathlib`.

I removed those lines as stale compatibility cleanup rather than because they affect supported runtimes. If minimizing the cleanup diff is preferred, I’m happy to retain the `__future__` imports and restore the guard as well.

oddstr13 (approved):

PR #1202, opened 2026-09-05
Use correct localization method

Rijul-A:
In #1195, it was wrongly believed that the localize string works for the
addon as well but it does not. Fix that, and in addition, use the string
from Kodi directly (so that we maintain fewer translations).

Added a comment to our strings.po to reflect that a current string
number is unused


oddstr13 (approved):
Based on the wiki page you linked https://kodi.wiki/view/Label_Parsing
it should likely have been `$ADDON[plugin.video.jellyfin 33263]`

But it also makes very much sense to piggy back off of Kodi's built in translations for basic stuff like a dialog close button.

PR #1204, opened 2026-09-05
Dump running threads and their stacks on service exit

oddstr13:
There have been several issues regarding Kodi hanging on exit.

This PR adds some more logging that hopefully sheds some light on what is going on in the background when this happens.

### DIFF
--- a/jellyfin_kodi/entrypoint/service.py
+++ b/jellyfin_kodi/entrypoint/service.py
@@ -6,6 +6,7 @@ import json
 import sys
 from datetime import datetime
 from importlib import reload
+import threading
 
 # Workaround for threads using datetime: _striptime is locked
 import _strptime  # noqa:F401
@@ -501,6 +502,9 @@ class Service(xbmc.Monitor):
         LOG.info("---[ objects reloaded ]")
 
     def shutdown(self):
+        LOG.debug("Running threads:")
+        for t in threading.enumerate():
+            LOG.debug("- %s: %s", t.name, t)
 
         LOG.info("---<[ EXITING ]")
         window("jellyfin_should_stop.bool", True)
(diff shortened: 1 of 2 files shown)


jellyfin_kodi/entrypoint/service.py
```diff
@@ -533,3 +537,7 @@ def shutdown(self):
             self.monitor.listener.stop()
 
         LOG.info("---<<<[ JELLYFIN ]")
```
Rijul-A, line 539: nit: Maybe this should be after the enumeration?

oddstr13: I think that is fine as is, this log line is just a flag, and the intent here is "These are the threads still running when this part of the code is done shutting down"

service.py
```diff
@@ -77,3 +77,12 @@ def run(self):
...
     LOG.info("--<[ service ]")
+
+    import traceback
+    import sys
+
+    LOG.debug("Running threads:")
+    for t in threading.enumerate():
+        LOG.debug("- %s: %s", t.name, t)
+        if t.ident is not None:
+            LOG.debug("".join(traceback.format_stack(sys._current_frames()[t.ident])))
```
Rijul-A: I got a KeyError with this one.

```
2026-09-08 14:41:43.210 T:1830171   error <general>: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                                   Error Type: <class 'KeyError'>
                                                   Error Contents: 140173145525376
                                                     File "/home/user/.kodi/addons/plugin.video.jellyfin/service.py", line 88, in <module>
                                                       LOG.debug("".join(traceback.format_stack(sys._current_frames()[t.ident])))
                                                                                                ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
                                                   KeyError: 140173145525376
```

```suggestion
            stack = sys._current_frames().get(t.ident)
            if stack is not None:
                LOG.debug("".join(traceback.format_stack(stack)))
```

oddstr13: Presumably a race condition where the thread exits during this for loop, before we get to it

oddstr13: I chose to do it a little differently, may or may not capture a few more stacks that would otherwise be gone.
KeyError should be gone at least.

Rijul-A (approved):

crobibero (approved):

PR #1209, opened 2026-09-07
Remove deprecated `translate_path`

Rijul-A:


oddstr13:
not doing a full review right now, but the http exception might need some more careful thinking to figure out exactly what to do (same with a fair bit of other exception stuff in the addon, that I haven't explicitly marked as deprecated).

translate_path is easy tho, as that one is just a wrapper around the underlying function now – it was added as a part of the py2->py3 transition, and is a easy search & replace.

oddstr13:
On that note, deprecated here is just an indication that someone has deemed that the particular piece of code should probably be removed in the future, there are several TODOs scattered around to similar effect.

For the HTTPException and other exception related stuff that's still left around, it's more of a "this doesn't quite look right, we should do something about it at some point", but what specifically is to be done (architectural choice) has yet to be decided.

Still, the effort is appreciated :)

Rijul-A:
Thank you for the feedback. If there is an issue where you discuss any of these (yet undecided) items, I'd love to be involved.

For now, I will revert my first commit and rename this PR to that effect.

jellyfin_kodi/helper/api.py
```diff
@@ -1,11 +1,11 @@
 # -*- coding: utf-8 -*-
 
 import json
+import xbmcvfs
```
oddstr13, line 5: Different import groups; builtin and 3rd party

```suggestion

import xbmcvfs
```

oddstr13 (approved):

oddstr13:
> Thank you for the feedback. If there is an issue where you discuss any of these (yet undecided) items, I'd love to be involved.

Unfortunately not, most of that is either in TODO/FIXME comments or just in my head.
There might be some issues for individual tasks, but most of it hasn't been discussed at GitHub. Some of it might have been discussed in chat (Matrix) at some point.

Issue #35, opened 2019-07-06
Video starts playing right away, but subtitles take a few seconds to load


oddstr13:
Seeking back to the start of the file once subtitles does load gives subtitles for the initial section also.

This may be an issue with how Jellyfin feeds us the data.

mothe:
I have this same issue, and it only happens when using Kodi + Jellyfin as a client. Web and Android app work fine. Video starts playing immediately but subs take ~5-10 seconds to load on a direct stream. If I pause the video and wait the 10 seconds, they will appear immediately after resuming.

Jellyfin 10.3.7 (Official Docker Container)
Kodi 18.3 and 18.4
Jellyfin Kodi Addon 0.2.1
Tested on: ASS and PGS subtitle filestypes in MKV containers. Both experience the delay.

Not sure what else I can share to help,

JustAMan:
I don't think this is a bug, more like how we expected it to behave... at least in web ui.
So that playback starts as fast as it can, and subtitles show when they're ready after extraction.

oddstr13:
That can't be expected behaviour. That's like audio starting to play a few seconds before the video is ready.

I would rather wait an extra couple of seconds for subtitles to load properly, or at least have a config switch for it, if this is indeed a intentional choice

leftl:
I am also having this issue, but only when using Kodi as a client. Web player and Android app work fine, as does direct playing the mkv container in Kodi via SMB. The loading delay only seems to happen with Jellyfin + Jellyfin-Kodi.
This would have to be an issue with the Jellyfin server though, right? Isn't this addon just building a local database of media served by Jellyfin, and Kodi is handling all of the streaming/decoding?

barolo:
Also having such issue, with latest version of everything. Furthermore unburned, srt subtitles will desync as tome point when transcoding [ unsure if it's related to hw accelerated transcoding, as I'm not using unaccelerated one at all ]

flashkiller999:
Can confirm the problem still exists.
Subtitle is delayed for about 5 seconds after video and audio already started playing.

Tested using a 1080p mkv with H264 (Bitrate: 2700 kb/s) and AAC, Subtitle format: SRT
Also tested using ASS/SSA: same behavior as with SRT.
Mode used: Direct play
Connection is a dedicated LAN connection (1 Gbits).

Problem does not occur when playing the file directly from a Samba share.
Problem does not occur on native Jellyfin-Android app and Webplayer. (Mode: Direct play)

[...]

PR #1210, opened 2026-09-08
Fix audio stream selection after playback starts

### LINKED ISSUE #1203: The Jellyfin-Kodi plugin keeps switching to audio stream 1, no matter what preferences are set in Jellyfin or Kodi
### User-Observed Behavior
* preferred audio language is set to English on both Jellyfin and Kodi
* playing a video file with two audio tracks: 1: de (default), 2: en will always select the German track
* switching the stream in Kodi works, but playing the same file a second time will play German again.

### Investigation
The Kodi log shows that the audio stream is first selected correctly and shortly after switched to German
```
00:11:30.111  Opening stream: 2 source: 256              <- Kodi selects English, correctly
00:11:32.101  jellyfin_kodi/player.py:199 Setting audio: 1 subs: None
00:11:32.161  Opening stream: 1 source: 256              <- add-on forces German
```
This seems to happen if inside `detect_audio_subs()` the call to `Player.GetProperties` fails. The `except` block sets `audio=0`. Later, we set `item["AudioStreamIndex"] = audio + 1`, which will in my case select stream 1, which is German.

### Attempted fix
Setting `audio = None` in the `except` block and later propagating that `None` into `item["AudioStreamIndex"]` solves the problem for me:
```diff
         try:  # Audio tracks
             audio = result["currentaudiostream"]["index"]
         except (KeyError, TypeError):
-            audio = 0
+            # Reading the current track failed, we don't know which track we're on.
+            audio = None
@@
-        item["AudioStreamIndex"] = audio + 1
+        item["AudioStreamIndex"] = audio + 1 if audio is not None else None
```
However, I don't know if this is the right way to fix this - I don't understand why this is set in the first place - it seems like this would always "re-set" the stream to whatever was already playing...?

### Bonus: AI-backed investigation why reading the audio stream fails in the first place
I'm not sure if this is helpful or even true, but I'll leave it here. Opus 5 claims that there is a race condition between the file starting to play and the streams being read:

```python
items = window("jellyfin_play.json")

while not items:
    if monitor.waitForAbort(2):   # only waits when the property is MISSING
        return
    items = window("jellyfin_play.json")
...
self.detect_audio_subs(item)      # so this runs immediately when it is present
```
> 
> `actions.py` has already written `jellyfin_play.json`, so the loop body never executes and there is no delay. `detect_audio_subs()` runs as soon as `onPlayBackStarted` fires — but `onPlayBackStarted` signals that playback was *initiated*, not that streams are open. `onAVStarted` is the callback for that.
```
00:11:30.111  audio thread created
00:11:30.203  jellyfin_kodi [ play ] kodi_id: 11042      <- add-on reads here
00:11:30.470  OnAVStarted                                <- streams actually ready
```

### Steps to Reproduce

1. Set the preferred audio language to English in both Jellyfin and Kodi
2. Play a video file with two audio tracks: 1: de (default), 2: en

**Current, buggy behavior:**
* Kodi plays the German track

**Expected behavior**
* Kodi honors the preferred language setting of both Jellyfin and Kodi itself in whatever order of preference - but since they are both set to English in my case, it plays the English track.

### Operating System

LibreELEC (Raspberry Pi 5)

### Jellyfin Server Version

Jellyfin 10.11.11

### Kodi Version

21

### Addon Mode

Addon-mode

### Logs

00:11:30.111  Opening stream: 2 source: 256              <- Kodi selects English, correctly
00:11:32.101  jellyfin_kodi/player.py:199 Setting audio: 1 subs: None
00:11:32.161  Opening stream: 1 source: 256              <- add-on forces German

### Server Logs

_No response_

### Additional Information

_No response_

Rijul-A:
This change fixes 2 closely related but distinct issues:
1. The false assumption that there is an arithmetic mapping between the server's media stream indexes and that of Kodi.
2. When playback starts, audio streams are not immediately available. If the current audio stream is fetched and not found, it gets overwritten with the 0th Kodi ordinal.

The fixes are:
1. Map Jellyfin audio stream indexes to Kodi stream ordinals instead of assuming the Jellyfin index corresponds to the Kodi stream position.
2. If Kodi responds with no available audio streams, do not overwrite the preferred media stream index in the item dictionary.

The test cases verify that:
1. The mapping is stored in the window
3. The look-up from Kodi ordinal to Jellyfin server index is detected
   correctly in `report_playback`
4. Given a request from the server to use a specific index, the mapping
   to the Kodi ordinal is accurately determined and used
5. Starting playback does not cause a loss of the requested Jellyfin
   server index and corresponding replacement with 0 or None

Fixes #1203

Reproduced with a DirectPlay MKV containing two external subtitles followed by video and two audio streams. Jellyfin selected English (`MediaStream.Index == 4`), while Kodi initially returned an empty `currentaudiostream`.

Before the change, the missing audio stream was treated as Kodi audio ordinal 0, overwriting the requested stream and resulting in Japanese audio. After the change, the unavailable stream is represented as `None`, the requested Jellyfin index 4 is preserved, and the `[3, 4]` audio-index mapping translates it to Kodi ordinal 1. Kodi subsequently reports ordinal 1 with language `en`.

Another good way to handle these would be to carve out `onAVStarted` separately; however, that is a bigger architectural change which will require a lot more testing to be certain.

### DIFF
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -106,6 +106,8 @@ class Player(xbmc.Player):
         window("jellyfin_play.json", items)
 
         self.set_item(current_file, item)
+        requested_audio = item["AudioStreamIndex"]
+        requested_subtitle = item["SubtitleStreamIndex"]
         # Detect current audio/subtitle state from Kodi player
         self.detect_audio_subs(item)
         data = {
@@ -141,7 +143,7 @@ class Player(xbmc.Player):
             return
 
         if item["PlayOption"] == "Addon":
-            self.set_audio_subs(item["AudioStreamIndex"], item["SubtitleStreamIndex"])
+            self.set_audio_subs(requested_audio, requested_subtitle)
 
     def set_item(self, file, item):
         """Set playback information."""
@@ -188,7 +190,7 @@ class Player(xbmc.Player):
         LOG.info("-->[ play/%s ] %s", item["Id"], item)
 
     def set_audio_subs(self, audio=None, subtitle=None):
-        if audio:
+        if audio is not None:
             audio = int(audio)
         if subtitle:
             subtitle = int(subtitle)
@@ -202,9 +204,14 @@ class Player(xbmc.Player):
 
             item = self.get_file_info(current_file)
             mapping = item["SubsMapping"]
+            kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
 
-            if audio and len(self.getAvailableAudioStreams()) > 1:
-                self.setAudioStream(audio - 1)
+            if (
+                audio is not None
+                and len(self.getAvailableAudioStreams()) > 1
+                and audio in kodi_audio_stream_indexes
+            ):
+                self.setAudioStream(kodi_audio_stream_indexes.index(audio))
 
             if subtitle is None or subtitle == -1:
                 self.showSubtitles(False)
@@ -237,7 +244,7 @@ class Player(xbmc.Player):
         try:  # Audio tracks
             audio = result["currentaudiostream"]["index"]
         except (KeyError, TypeError):
-            audio = 0
+            audio = None
 
         try:  # Subtitles tracks
             subs = result["currentsubtitle"]["index"]
@@ -249,7 +256,12 @@ class Player(xbmc.Player):
         except (KeyError, TypeError):
             subs_enabled = False
 
-        item["AudioStreamIndex"] = audio + 1
+        kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
+        item["AudioStreamIndex"] = (
+            kodi_audio_stream_indexes[audio]
+            if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes)
+            else None
+        )
 
         if not subs_enabled or not len(self.getAvailableSubtitleStreams()):
             item["SubtitleStreamIndex"] = None
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -257,11 +257,9 @@ class Player(xbmc.Player):
             subs_enabled = False
 
         kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
-        item["AudioStreamIndex"] = (
-            kodi_audio_stream_indexes[audio]
-            if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes)
-            else None
-        )
+        item["AudioStreamIndex"] = None
+        if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes):
+            item["AudioStreamIndex"] = kodi_audio_stream_indexes[audio]
 
         if not subs_enabled or not len(self.getAvailableSubtitleStreams()):
             item["SubtitleStreamIndex"] = None
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -106,8 +106,6 @@ class Player(xbmc.Player):
         window("jellyfin_play.json", items)
 
         self.set_item(current_file, item)
-        requested_audio = item["AudioStreamIndex"]
-        requested_subtitle = item["SubtitleStreamIndex"]
         # Detect current audio/subtitle state from Kodi player
         self.detect_audio_subs(item)
         data = {
@@ -143,7 +141,7 @@ class Player(xbmc.Player):
             return
 
         if item["PlayOption"] == "Addon":
-            self.set_audio_subs(requested_audio, requested_subtitle)
+            self.set_audio_subs(item["AudioStreamIndex"], item["SubtitleStreamIndex"])
 
     def set_item(self, file, item):
         """Set playback information."""
@@ -256,8 +254,10 @@ class Player(xbmc.Player):
         except (KeyError, TypeError):
             subs_enabled = False
 
+        # When playback is started, the audiostream is not available
+        # In such a case, audio is None and we must not overwrite the
+        # item level value
         kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
-        item["AudioStreamIndex"] = None
         if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes):
             item["AudioStreamIndex"] = kodi_audio_stream_indexes[audio]
 
--- a/tests/test_player.py
+++ b/tests/test_player.py
@@ -7,9 +7,8 @@ from jellyfin_kodi.helper import playutils as playutils_module
 def test_audio_stream_indexes_are_stored_for_playback(monkeypatch):
     # Validate that KodiAudioStreamIndexes is set, given the
     # Audio streams in the source file
-    class ApiClient:
-        class config:
-            data = {"auth.token": ""}
+    api_client = Mock()
+    api_client.config.data = {"auth.token": ""}
 
     item = {
         "Id": "item",
@@ -42,19 +41,24 @@ def test_audio_stream_indexes_are_stored_for_playback(monkeypatch):
 
     monkeypatch.setattr(playutils_module, "window", fake_window)
     monkeypatch.setattr(playutils_module.client, "get_device_id", lambda: "device")
+
+    class Api:
+        def get_file_path(self, _):
+            return "movie"
+
     monkeypatch.setattr(
         playutils_module.api,
         "API",
-        lambda *_args, **_kwargs: Mock(get_file_path=Mock(return_value="movie")),
+        lambda *_args, **_kwargs: Api(),
     )
 
     play_utils = playutils_module.PlayUtils(
         item,
         server="http://server",
-        api_client=ApiClient,
+        api_client=api_client,
     )
     play_utils.info["Path"] = "movie"
-    play_utils.direct_url = Mock()
+    play_utils.direct_url = lambda _source: None
 
     play_utils.get(source)
     playutils_module.set_properties(item, "DirectStream", "server")
@@ -64,17 +68,15 @@ def test_audio_stream_indexes_are_stored_for_playback(monkeypatch):
 
 
 def test_detect_audio_stream_maps_kodi_ordinal_to_jellyfin_index(monkeypatch):
-    # Verify that the Kodi audio ordinal that is currently set
-    # maps back correctly to the AudioStreamIndex on the server
-    # This is used in report_playback to detect which audio
-    # is currently playing
+    # Kodi reports audio stream ordinals
+    # Jellyfin uses MediaStream indexes.
     player = object.__new__(player_module.Player)
     item = {
         "KodiAudioStreamIndexes": [3, 4],
         "SubsMapping": {},
     }
 
-    class JsonRpc:
+    class JSONRPC:
         def __init__(self, _):
             pass
 
@@ -87,7 +89,7 @@ def test_detect_audio_stream_maps_kodi_ordinal_to_jellyfin_index(monkeypatch):
                 }
             }
 
-    monkeypatch.setattr(player_module, "JSONRPC", JsonRpc)
+    monkeypatch.setattr(player_module, "JSONRPC", JSONRPC)
 
     player.detect_audio_subs(item)
 
@@ -95,8 +97,8 @@ def test_detect_audio_stream_maps_kodi_ordinal_to_jellyfin_index(monkeypatch):
 
 
 def test_set_audio_stream_maps_jellyfin_index_to_kodi_ordinal():
-    # Verify that calling set_audio_subs with a Jellyfin
-    # index maps to the correct Kodi ordinal
+    # Jellyfin uses MediaStream indexes
+    # Kodi expects audio ordinals.
     player = object.__new__(player_module.Player)
     item = {
         "KodiAudioStreamIndexes": [3, 4],
@@ -108,7 +110,6 @@ def test_set_audio_stream_maps_jellyfin_index_to_kodi_ordinal():
     player.is_playing_file = lambda _: True
     player.getAvailableAudioStreams = lambda: [{}, {}]
     player.setAudioStream = Mock()
-    player.showSubtitles = Mock()
 
     player.set_audio_subs(audio=4)
 
@@ -116,9 +117,9 @@ def test_set_audio_stream_maps_jellyfin_index_to_kodi_ordinal():
 
 
 def test_playback_started_preserves_requested_audio_stream(monkeypatch):
-    # Assert that starting playback does not set the
-    # requested Jellyfin server audio index to None or
-    # even 0
+    # Kodi may not report the current audio stream
+    # immediately after playback starts. Preserve
+    # the requested stream while detecting it.
     player = object.__new__(player_module.Player)
 
     server = Mock()
@@ -152,7 +153,7 @@ def test_playback_started_preserves_requested_audio_stream(monkeypatch):
     monkeypatch.setattr(player_module.xbmc, "Monitor", Mock(return_value=monitor))
     monkeypatch.setattr(player_module, "settings", Mock(return_value=False))
 
-    class JsonRpc:
+    class JSONRPC:
         def __init__(self, _):
             pass
 
@@ -165,7 +166,7 @@ def test_playback_started_preserves_requested_audio_stream(monkeypatch):
                 }
             }
 
-    monkeypatch.setattr(player_module, "JSONRPC", JsonRpc)
+    monkeypatch.setattr(player_module, "JSONRPC", JSONRPC)
 
     def fake_window(name, value=None, **kwargs):
         if name == "jellyfin_play.json" and value is None:
--- a/tests/test_player.py
+++ b/tests/test_player.py
@@ -176,5 +176,5 @@ def test_playback_started_preserves_requested_audio_stream(monkeypatch):
 
     player.onPlayBackStarted()
 
-    assert item["AudioStreamIndex"] is None
+    assert item["AudioStreamIndex"] == 4
     player.set_audio_subs.assert_called_once_with(4, None)
--- a/tests/test_player.py
+++ b/tests/test_player.py
@@ -1,0 +1,179 @@
+from unittest.mock import Mock
+
+import jellyfin_kodi.player as player_module
+from jellyfin_kodi.helper import playutils as playutils_module
+
+
+def test_audio_stream_indexes_are_stored_for_playback(monkeypatch):
+    # Validate that KodiAudioStreamIndexes is set, given the
+    # Audio streams in the source file
+    class ApiClient:
+        class config:
+            data = {"auth.token": ""}
+
+    item = {
+        "Id": "item",
+        "Type": "Movie",
+        "MediaType": "Video",
+    }
+    source = {
+        "Id": "source",
+        "Path": "movie",
+        "SupportsDirectPlay": False,
+        "SupportsDirectStream": True,
+        "MediaStreams": [
+            {"Index": 0, "Type": "Subtitle"},
+            {"Index": 1, "Type": "Subtitle"},
+            {"Index": 2, "Type": "Video"},
+            {"Index": 3, "Type": "Audio"},
+            {"Index": 4, "Type": "Audio"},
+        ],
+        "DefaultAudioStreamIndex": 4,
+        "DefaultSubtitleStreamIndex": None,
+    }
+
+    windows = {}
+
+    def fake_window(name, value=None, **kwargs):
+        if name == "jellyfin_play.json":
+            if value is None:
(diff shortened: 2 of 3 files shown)


Rijul-A:
I don't know why Sonar is reporting 4 new issues; I already fixed them with a force push.

Rijul-A:
I reproduced the originally reported issue on master and tested it out with this patch applied.

The media streams, as reported by the server, are listed below. Observe how 4 (English) is chosen based on the user's settings.

```json
{
  "DefaultAudioStreamIndex": 4,
  "DefaultSubtitleStreamIndex": 1,
  "MediaStreams": [
    { "Index": 0, "Type": "Subtitle", "Language": "eng", "IsExternal": true },
    { "Index": 1, "Type": "Subtitle", "Language": "eng", "IsExternal": true },
    { "Index": 2, "Type": "Video" },
    { "Index": 3, "Type": "Audio", "Language": "jpn", "IsDefault": true },
    { "Index": 4, "Type": "Audio", "Language": "eng", "IsDefault": false }
  ]
}
```

In Kodi, the two audio tracks are exposed as ordinals 0 and 1 for Japanese and English respectively.

I added some verbose logging to demonstrate that the server reports a request of 4, but the data is not yet available. The fallback mechanism changes that to Kodi ordinal 0 which becomes server stream 1 (Japanese).

```text
TEST requested before detection: audio=4
TEST Player.GetProperties result: {'currentaudiostream': {}, 'currentsubtitle': {}, 'subtitleenabled': True}
TEST Kodi current audio ordinal: 0
TEST after detection: audio=1
...
Setting audio: 1 subs: 0
```

Kodi then reported this audiostream.

```py
{
    'index': 0,
    'language': 'ja',
    'isdefault': True,
    'name': 'E-AC-3 5.1'
}
```

With this patch applied and the verbose logging retained, the None value is not applied and the original stored value is retained to be applied once Kodi loads the streams.

```text
TEST requested before detection: audio=4
TEST Player.GetProperties result: {'currentaudiostream': {}, 'currentsubtitle': {}, 'subtitleenabled': True}
TEST Kodi current audio ordinal: None
TEST after detection: audio=None
...
Setting audio: 4 subs: 0
TEST KodiAudioStreamIndexes: [3, 4]
```

Kodi then reported the correct audiostream.

```py
{
    'index': 1,
    'language': 'en',
    'isdefault': False,
    'name': 'E-AC-3 5.1'
}
```

For this test item, the audio mapping is `[3, 4]`, so Jellyfin stream `4` maps to Kodi ordinal `1`.

This reproduction also shows two separate assumptions in the existing path. The immediate #1203 failure is caused by treating a missing `currentaudiostream` as ordinal 0 and overwriting the requested stream. Separately, the preserved Jellyfin stream index cannot be passed to Kodi as `audio - 1`: for this item Jellyfin audio indexes are `[3, 4]`, while Kodi exposes them as ordinals `[0, 1]`. The patch addresses both so the originally requested stream can be preserved and then applied correctly.

Rijul-A:
Subtitle stream handling has a similar index/ordinal mismatch, but that path has additional mapping behavior around external subtitles. I’m keeping that out of this PR and will handle it separately.

oddstr13:
I haven't done a review, just briefly glanced at the code tab.
Any AI assistance with creating those tests, or are they hand-made?
I've seen a fair bit of bad AI tests in relation to the PRs I just closed, and this is important information determining how much scrutiny is required during review.

Rijul-A:
> I haven't done a review, just briefly glanced at the code tab. Any AI assistance with creating those tests, or are they hand-made? I've seen a fair bit of bad AI tests in relation to the PRs I just closed, and this is important information determining how much scrutiny is required during review.

Tests are now hand made.

oddstr13:
I'm not saying I require tests to be fully hand-made, but generated tests really do need a few extra review passes to ensure they actually do what they say on the tin.
Thanks for going over them once! :)

jellyfin_kodi/player.py
```diff
@@ -248,7 +256,12 @@ def detect_audio_subs(self, item):
         except (KeyError, TypeError):
             subs_enabled = False
 
-        item["AudioStreamIndex"] = audio + 1
+        kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
+        item["AudioStreamIndex"] = (
+            kodi_audio_stream_indexes[audio]
+            if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes)
+            else None
+        )
```
oddstr13: Assuming None is expected here, and that the absence of the item wouldn't be more appropriate in this particular location?
```suggestion
        item["AudioStreamIndex"] = None
        if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes):
            item["AudioStreamIndex"] = kodi_audio_stream_indexes[audio]
```

A normal if reads better than a ternary statement here (we're not code-golfing)

jellyfin_kodi/player.py
```diff
@@ -140,7 +143,7 @@ def onPlayBackStarted(self):
             return
 
         if item["PlayOption"] == "Addon":
-            self.set_audio_subs(item["AudioStreamIndex"], item["SubtitleStreamIndex"])
+            self.set_audio_subs(requested_audio, requested_subtitle)
```
oddstr13: Any particular reason, except for it looking a tiny bit nicer?
Looks like a leftovers from  a different implementation attempt.

Edit: diff view was supposed to also include the above assignments of those variables

oddstr13 (changes requested):
Needs independent runtime testing «I see no change» is an acceptable result assuming you don't have a good way of reproducing the issue.
I'll get to it at some point if nobody does it before me.

The tests need a full review, I've only taken a quick look but I'm not entirely sure of how much work they actually do due to the extensive monkeypatching. (coverage reports would be helpful here; TODO)

jellyfin_kodi/player.py
Rijul-A: The `detect_audio_subs` overwrites these values so we must save them beforehand.

Rijul-A:
> The tests need a full review, I've only taken a quick look but I'm not entirely sure of how much work they actually do due to the extensive monkeypatching. (coverage reports would be helpful here; TODO)

IMO the tests cover exactly everything listed in the description and the example. The monekypatching is performed to focus solely on the objective of the tests and skip the surrounding pieces but I understand your concern.

1. Jellyfin provides audio streams with different indexes => They are mapped to Kodi 0-based indexes and the mapping is saved into the item.
2. The mapping so-generated is used correctly to forward / reverse lookup the index.
3. If Kodi reports no active audio streams, user preferred stream indexes aren't lost.

This will help create a 10s test file but I don't know whether the server will follow the same indexes for the stream as my example. I checked the result via `mediainfo` but the command is AI generated.

EDIT: reordered the streams to put subtitles first

```bash
printf '%s\n' \
  '1' \
  '00:00:01,000 --> 00:00:03,000' \
  '[Forced subtitle]' \
  > forced.srt

printf '%s\n' \
  '1' \
  '00:00:01,000 --> 00:00:03,000' \
  'Regular English subtitle' \
  > english.srt

ffmpeg \
  -f lavfi -i 'testsrc2=size=3840x1606:rate=24000/1001' \
  -f lavfi -i 'anullsrc=r=48000:cl=5.1' \
  -f lavfi -i 'anullsrc=r=48000:cl=5.1' \
  -i forced.srt \
  -i english.srt \
  -map 3:0 \
  -map 4:0 \
  -map 0:v:0 \
  -map 1:a:0 \
  -map 2:a:0 \
  -t 10 \
  -c:v libx265 \
  -pix_fmt yuv420p10le \
  -preset medium \
  -crf 28 \
  -color_range tv \
  -color_primaries bt2020 \
  -color_trc smpte2084 \
  -colorspace bt2020nc \
  -x265-params 'hdr10=1:repeat-headers=1:master-display=G(8500,39850)B(6550,2300)R(35400,14600)WP(15635,16450)L(10000000,1):max-cll=996,817' \
  -c:a eac3 \
  -b:a 640k \
  -ar 48000 \
  -ac 6 \
  -c:s copy \
  -metadata:s:v:0 language=und \
  -metadata:s:a:0 language=jpn \
  -metadata:s:a:0 title='E-AC-3 5.1' \
  -metadata:s:a:1 language=eng \
  -metadata:s:a:1 title='E-AC-3 5.1' \
  -metadata:s:s:0 language=eng \
  -metadata:s:s:0 title='English (forced - SRT)' \
  -metadata:s:s:1 language=eng \
  -metadata:s:s:1 title='English (SRT)' \
  -disposition:v:0 default \
  -disposition:a:0 default \
  -disposition:a:1 0 \
  -disposition:s:0 forced \
  -disposition:s:1 0 \
  test-file.mkv
```

jellyfin_kodi/player.py
oddstr13: hmm, that begs the question of whether that function is *supposed* to do what you're fixing here?
Is this function getting obsoleted by your work here (and the mentioned TBD subtitle)?
Is the function not being used appropriately?

I see that it is getting called from two locations; `onPlaybackStarted` and `report_playback`, the latter of which gets executed multiple times per playback.

I haven't stepped thru the function and its use to fully understand what it does.

oddstr13:
My preference for tests would be to monkeypatch only the things that are required to get the tests running, and to otherwise do as much end-to-end testing as is natural for the functionality under test. One example of this would be `kodi_version`, where the patch is included in the function from the beginning of.

I've been wanting to mock API responses (monkey-patching requests or the layer just above it) and Kodi UI stuff, but that's way out of scope for this PR.

If you think the tests properly demonstrate the issue to be fixed, and that you're confident they would be useful for catching potential future errors in the handling of audio tracks, then that's fine and we'll leave them as is (pending proper review, as I do want to ensure tests actually do work – a false sense of safety is worse than knowing the railing doesn't exist…).

jellyfin_kodi/player.py
Rijul-A: Good point. I modified the function itself so that it no longer overwrites `item["AudioStreamIndex"]` if Kodi cannot provide the current audio stream.

This modification will work with `report_playback` as well; if Kodi reports that the audiostream is not yet available, the client will report the existing value to the server unchanged.

tests/test_player.py
```diff
@@ -0,0 +1,180 @@
...
+            {"Index": 3, "Type": "Audio"},
+            {"Index": 4, "Type": "Audio"},
+        ],
+        "DefaultAudioStreamIndex": 4,
+        "DefaultSubtitleStreamIndex": None,
+    }
+
+    windows = {}
+
+    def fake_window(name, value=None, **kwargs):
```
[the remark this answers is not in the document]
Rijul-A: It is a test but I have added explicit `return None` statements.

tests/test_player.py
```diff
@@ -0,0 +1,180 @@
...
+                "result": {
+                    "currentaudiostream": {},
+                    "currentsubtitle": {},
+                    "subtitleenabled": True,
+                }
+            }
+
+    monkeypatch.setattr(player_module, "JSONRPC", JSONRPC)
+
+    def fake_window(name, value=None, **kwargs):
```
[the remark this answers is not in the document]
Rijul-A: Added

jellyfin_kodi/player.py
```diff
@@ -248,7 +254,12 @@ def detect_audio_subs(self, item):
         except (KeyError, TypeError):
             subs_enabled = False
 
-        item["AudioStreamIndex"] = audio + 1
+        # When playback is started, the audiostream is not available
+        # In such a case, audio is None and we must not overwrite the
+        # item level value
```
oddstr13, line 259: Not of importance for this PR, but this sounds like it might describe the behavior of subtitles in #35, anything remotely similar happening for subs somewhere?

tests/test_player.py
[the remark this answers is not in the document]
oddstr13: "Used in tests" is one of the options for dismissing, so either is likely fine here

jellyfin_kodi/player.py
Rijul-A: I have seen the issue too. Both audio and subtitle are not immediately available so it is possibly related. I will investigate further.

oddstr13:
Verified that added tests indeed fail pre patch :ballot_box_with_check:

tests/test_player.py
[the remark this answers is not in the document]
Rijul-A: I don't think that button was available to me

jellyfin_kodi/player.py
Rijul-A: I found a few sample files in my library that are thus affected. I will investigate using them and submit the subtitle fix separately.

oddstr13:
I had audio seemingly cut out for half a sec just after starting the file on a multi-audiotrack episode – seems similar to what is happening with subtitles.

Presumably starting with whatever Kodi thinks is the default, and then switching (in this case to the same one) once some code runs and specifies which track to use.

I need to locate more dual audio files in my collection for better testing this.

Might be worth noting that I don't have a specific language set as preferred, due to the nature of the media I watch (I want original language audio most of the time, with English subs if needed).

Likely not this PR causing the effect, just that it actually specifying a track now is exposing it.


More testing needed.

Rijul-A:
Given the verified finding in #1203 about `onAVStarted`, I believe this patch may work better. However, it does look like a bigger architectural change so I haven't committed it.

I tested it out and I did not get the half a sec lag but I also couldn't reproduce it on 9eea46535d7baaf21dd2c3fe5e5196a0c702b438.

```patch
diff --git a/jellyfin_kodi/helper/playutils.py b/jellyfin_kodi/helper/playutils.py
index 72db42bd..3f3129a2 100644
--- a/jellyfin_kodi/helper/playutils.py
+++ b/jellyfin_kodi/helper/playutils.py
@@ -221,9 +221,17 @@ class PlayUtils(object):
             LOG.info("--[ transcode ]")
             self.transcode(source, audio, subtitle)
 
-        self.info["AudioStreamIndex"] = self.info.get("AudioStreamIndex") or source.get(
-            "DefaultAudioStreamIndex"
-        )
+        if self.info.get("AudioStreamIndex") is None:
+            self.info["AudioStreamIndex"] = source.get("DefaultAudioStreamIndex")
+
+        if self.info.get("Method") == "Transcode":
+            # When transcoding there is only one audio
+            selected_audio = self.info["AudioStreamIndex"]
+            if selected_audio is not None and selected_audio >= 0:
+                self.info["KodiAudioStreamIndexes"] = [selected_audio]
+            else:
+                self.info["KodiAudioStreamIndexes"] = []
+
         self.info["SubtitleStreamIndex"] = self.info.get(
             "SubtitleStreamIndex"
         ) or source.get("DefaultSubtitleStreamIndex")
diff --git a/jellyfin_kodi/player.py b/jellyfin_kodi/player.py
index 2c60fcf8..814f9bb1 100644
--- a/jellyfin_kodi/player.py
+++ b/jellyfin_kodi/player.py
@@ -49,38 +49,24 @@ class Player(xbmc.Player):
         return file in self.played
 
     def onPlayBackStarted(self):
-        """We may need to wait for info to be set in kodi monitor.
-        Accounts for scenario where Kodi starts playback and exits immediately.
-        First, ensure previous playback terminated correctly in Jellyfin.
-        """
-
+        """Reset lifecycle state before Kodi starts its AV streams."""
+        LOG.debug("Kodi onPlayBackStarted received; resetting playback state")
         self.stop_playback()
         self._reset_state(restart=True)
-        count = 0
-        monitor = xbmc.Monitor()
-
-        try:
-            current_file = self.getPlayingFile()
 
-        except Exception:
-
-            while count < 5:
-                try:
-                    current_file = self.getPlayingFile()
-                    count = 0
-                    break
-                except Exception:
-                    count += 1
+    def onAVStarted(self):
+        """Initialize Jellyfin playback once Kodi has usable AV streams."""
+        LOG.debug("Kodi onAVStarted received")
+        current_file = self.get_playing_file()
 
-                if monitor.waitForAbort(1):
-                    return
-            else:
-                LOG.info("Cancel playback report")
+        if not current_file:
+            LOG.info("Cancel playback report")
 
-                return
+            return
 
         items = window("jellyfin_play.json")
-        item = None
+        count = 0
+        monitor = xbmc.Monitor()
 
         while not items:
 
@@ -102,12 +88,27 @@ class Player(xbmc.Player):
                 break
         else:
             item = items.pop(0)
+        LOG.debug(
+            "AV-ready item %s: option=%s, requested audio=%s, requested subtitle=%s",
+            item["Id"],
+            item["PlayOption"],
+            item["AudioStreamIndex"],
+            item["SubtitleStreamIndex"],
+        )
 
         window("jellyfin_play.json", items)
-
         self.set_item(current_file, item)
-        # Detect current audio/subtitle state from Kodi player
-        self.detect_audio_subs(item)
+
+        if not self.is_playing_file(current_file):
+            return
+
+        if item["PlayOption"] == "Addon":
+            self.set_audio_subs(
+                item["AudioStreamIndex"], item["SubtitleStreamIndex"]
+            )
+        else:
+            self.detect_audio_subs(item)
+
         data = {
             "QueueableMediaTypes": "Video,Audio",
             "CanSeek": True,
@@ -137,12 +138,6 @@ class Player(xbmc.Player):
             except Exception:
                 pass  # Player may not be ready yet
 
-        if monitor.waitForAbort(2):
-            return
-
-        if item["PlayOption"] == "Addon":
-            self.set_audio_subs(item["AudioStreamIndex"], item["SubtitleStreamIndex"])
-
     def set_item(self, file, item):
         """Set playback information."""
         try:
@@ -203,20 +198,35 @@ class Player(xbmc.Player):
             item = self.get_file_info(current_file)
             mapping = item["SubsMapping"]
             kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
+            available_audio_streams = self.getAvailableAudioStreams()
+            LOG.debug(
+                "Audio selection: requested Jellyfin stream=%s, Kodi mapping=%s, available=%d",
+                audio,
+                kodi_audio_stream_indexes,
+                len(available_audio_streams),
+            )
 
             if (
                 audio is not None
-                and len(self.getAvailableAudioStreams()) > 1
+                and len(available_audio_streams) > 1
                 and audio in kodi_audio_stream_indexes
             ):
-                self.setAudioStream(kodi_audio_stream_indexes.index(audio))
+                kodi_audio_stream = kodi_audio_stream_indexes.index(audio)
+                LOG.debug(
+                    "Selecting Kodi audio stream %d for Jellyfin stream %d",
+                    kodi_audio_stream,
+                    audio,
+                )
+                self.setAudioStream(kodi_audio_stream)
+            else:
+                LOG.debug("No Kodi audio stream override required or available")
 
             if subtitle is None or subtitle == -1:
                 self.showSubtitles(False)
 
                 return
 
-            tracks = len(self.getAvailableAudioStreams())
+            tracks = len(available_audio_streams)
 
             if mapping:
                 for index in mapping:
@@ -253,13 +263,26 @@ class Player(xbmc.Player):
             subs_enabled = result["subtitleenabled"]
         except (KeyError, TypeError):
             subs_enabled = False
+        LOG.debug(
+            "Detected Kodi streams: audio=%s, subtitle=%s, subtitles enabled=%s",
+            audio,
+            subs,
+            subs_enabled,
+        )
 
         # When playback is started, the audiostream is not available
         # In such a case, audio is None and we must not overwrite the
-        # item level value
         kodi_audio_stream_indexes = item.get("KodiAudioStreamIndexes") or []
-        if audio is not None and 0 <= audio < len(kodi_audio_stream_indexes):
-            item["AudioStreamIndex"] = kodi_audio_stream_indexes[audio]
+        if audio is not None:
+            if 0 <= audio < len(kodi_audio_stream_indexes):
+                item["AudioStreamIndex"] = kodi_audio_stream_indexes[audio]
+            else:
+                LOG.warning(
+                    "Kodi audio stream %d has no Jellyfin mapping in %s",
+                    audio,
+                    kodi_audio_stream_indexes,
+                )
+                item["AudioStreamIndex"] = None
 
         if not subs_enabled or not len(self.getAvailableSubtitleStreams()):
             item["SubtitleStreamIndex"] = None
```

Tests are not updated for the above patch.

Rijul-A:
Perhaps the time gap is caused by the fact that Kodi first uses its internal settings to determine the preferred audio stream and then the callback is fired which permits Jellyfin to change it.

oddstr13:
> Perhaps the time gap is caused by the fact that Kodi first uses its internal settings to determine the preferred audio stream and then the callback is fired which permits Jellyfin to change it.

Yes, I think it is very likely something along those lines, both for audio and subtitles.

Would probably be best to avoid "changing" them if we can detect that the right stream is already selected. Maybe even an advanced setting to allow Kodi to do the selection and get out of the way completely (except ensuring that external subs are present).

All of that is out of scope for this PR tho.

oddstr13 (approved):
Except for the audio-blipp (of which fixing I consider out of scope for this PR), I don't see any negative changes in my own setup.

PR #1236, opened 2026-09-18
Fix skip dialog layout for localized labels

Hunduster:
The skip dialog currently uses auto-sized buttons inside a fixed-width 460px container. This can cause the layout to overflow when localized labels are significantly longer than their English equivalents.

This became particularly noticeable with German localization after the skip dialog was localized in 2.2.0.

For example:

- `Recap` → `Zusammenfassung`
- `Skip {0} ({1})` → `Überspringe {0} ({1})`
- `Close` → `Schließen`

A label such as `Überspringe Zusammenfassung (1m 5s)` makes the skip button grow beyond the available space. As a result, the Close button is pushed outside the visible dialog area. It can still be focused using the remote, but is partially or completely hidden.

### Solution

This change removes the fixed background and accent bar and uses a wider, right-aligned grouplist instead.

The existing auto-sized Skip and Close buttons are left unchanged. This allows the buttons to adapt naturally to different label lengths while keeping the right edge of the controls at the same screen position.

Unlike simply increasing the fixed background width, this also avoids excessive empty background space for shorter labels.

### Tested

Tested on Kodi with Jellyfin for Kodi 2.2.0 using German localization with different segment label lengths:

- Introduction: `Überspringe Vorspann`
- Recap: `Überspringe Zusammenfassung`

Both the Skip and Close buttons remain fully visible and correctly aligned.


oddstr13:
Please try #1191

Hunduster:
> Please try #1191

Thanks for pointing me to #1191.

I had not seen that PR before opening mine, so the overlap was unintentional.

I also want to say that I appreciate the work that went into #1191. It is clearly a more comprehensive redesign of the skip dialog.

That said, I personally prefer the approach in this PR for a few reasons:

- the buttons keep their automatic width and adapt directly to the localized label length
- there is no fixed background area that has to be sized for the longest possible translation
- the transparent layout feels less intrusive during playback
- no thumbnail/artwork is added, which keeps the overlay focused on the actual action

For me, especially on a TV, this results in a simpler and more elegant UI. The dialog only takes up as much space as the current labels require, so shorter strings such as `Überspringe Vorspann` stay compact while longer ones such as `Überspringe Zusammenfassung` still fit correctly.

So while I understand the direction of #1191, I would prefer to keep this PR as the smaller layout-focused fix rather than switch to the redesigned dialog.

Of course, I am happy to adjust anything in this implementation if there are technical concerns with the dynamic layout itself.

oddstr13:
#1191 is just pending some minor tweaks and a quick test by someone who use the feature, so anything here would have to go on top of that.

Please take note of the before and after screenshots provided in that PR, those are extremely helpful when judging UI changes.

One concern with a dynamic layout, would be that buttons can change positions, which is often not desirable.

As for not wanting the image, that is completely fair, and I do think a settings toggle would be in order for that.

Hunduster:
Thanks, that makes sense.

I understand that #1191 is intended to become the new baseline for the dialog, so any further changes should be based on that once it is merged.

Regarding the dynamic layout: the implementation in this PR is right-aligned to a fixed right edge (x=1880). This means the Close button remains anchored at the same position. Only the left edge of the Skip button moves depending on the localized label length, as the group expands to the left.

I agree that controls unexpectedly moving around can be undesirable, but in this case I actually found that behavior quite natural during testing, since the primary and secondary actions remain in the same order and the right edge stays fixed.

And yes, a setting to disable the image in #1191 sounds like a very good compromise to me. I personally prefer the minimal overlay without artwork, especially during playback, but making that optional would accommodate both approaches nicely.

I'll also keep the before/after screenshot point in mind for future UI changes. I agree that would have made this PR easier to evaluate.

Rijul-A:
> And yes, a setting to disable the image in #1191 sounds like a very good compromise to me. I personally prefer the minimal overlay without artwork, especially during playback, but making that optional would accommodate both approaches nicely.

Thanks; I will add this in my next PR as requested.

PR #1237, opened 2026-09-18
Keep the server-provided path intact for live TV transcoding

rubdos:
Rewriting the transcoding base path from "stream" to "live" made live TV playback request endpoints such as live.ts, which current Jellyfin servers do not serve, resulting in 404 on every live TV play attempt.

Current servers emit /videos/{id}/stream.ts (or /stream) and serve the opened live stream from that URL, so keep the server-provided path for live TV and only apply the HLS master rewrite for non-live content.

This was debugged and patched with an LLM (together with #1238, because I had both issues simultaneously), This is tested and verified against Jellyfin 10.11.8. The LLM (GLM 5.3 flash) claims that this is most probably an artifact from Emby times.

I'll be happy to dive deeper into this manually if this doesn't seem right. Out of both my patches, this seems to be the most straight-forwardly correct to me.


oddstr13:
I'd like to know why we're doing that replace in the first place before changing this any.

I'm seeing `/Videos/{itemId}/live.m3u8` in the openapi specs for versions `10.7.0` thru `10.11.11`, but not in `12.*`, a link to the PR changing this upstream is desirable.

Depending on what actually got changed, we may need to do some magic to keep backwards compat.

Rijul-A:
https://github.com/jellyfin/jellyfin/pull/16715

Issue #983, opened 2025-03-09
The path replacement feature does not work properly when using WebDAV with non-English characters in the path.


xspeed1989:
### Summary

The path replacement feature does not work properly when using WebDAV with non-English characters in the path.

### Steps to Reproduce

Set up a WebDAV server and configure the path replacement feature in Jellyfin/Kodi to use WebDAV paths.

### Operating System

Android 14

### Jellyfin Server Version

10.10.5

### Kodi Version

21.2

### Addon Mode

Direct-path mode ('native')

### Logs

2025-03-09 22:12:25.636 T:25101    info <general>: Creating InputStream
2025-03-09 22:12:25.636 T:25101   debug <general>: CFileCache::Open - <davs://hide:12345/jellyfin_media/movie/源代码 (2011)> opening
2025-03-09 22:12:25.637 T:25101   debug <general>: CurlFile::Open - <https://hide:12345/jellyfin_media/movie/源代码 (2011)>
2025-03-09 22:12:25.637 T:25101   error <general>: CCurlFile::CReadState::FillBuffer - (0xb400007338b47cb0) Failed: URL using bad/illegal format or missing URL(3)
2025-03-09 22:12:25.637 T:25101   error <general>: CCurlFile::Open - <https://hide:12345/jellyfin_media/movie/源代码 (2011)> Failed with code 0:
                                                   
2025-03-09 22:12:25.637 T:25101   error <general>: CFileCache::Open - <davs://hide:12345/jellyfin_media/movie/源代码 (2011)> failed to open

PR #1238, opened 2026-09-18
Set app identity on every Jellyfin client construction

rubdos:
Jellyfin clients constructed outside of the login flow (e.g. the playback entrypoint) carried no `device_id`, so the HTTP layer fell back to placeholder values ("Unknown Device id") in the Authorization header. The server echoes that DeviceId into streaming URLs, and values with unresolved spaces make the request line illegal (I'd say similar to #983), breaking live TV playback.

Should also fix stale "Unknown Device" entries appearing on the server's device list.

This was debugged and patched with an LLM (together with #1237), because I had both issues simultaneously). This is tested and verified against Jellyfin 10.11.8. I don't necessarily like the import workaround it did, but I'll be happy to patch that up somehow.


oddstr13:
I'm not sure of what's going on here, but whatever it is, this is not the right way of fixing it.

Please fill out a proper issue with reproduction steps (how to trigger the issue).

I don't see any `Unknown Device` in my server's device list, but I may not be going thru the same menus and settings as you are, and not hitting the same code paths.

